PR #1, opened 2019-11-25
WIP - outer joins

tpolecat:
A few trivial changes to start out:

- Removed extra spaces from generated SQL.
- Picked a better example for the `recursive query (1)` test to show that it works with countries that have more than one city.
- Fixed handling of empty resultsets so we get "No match" rather than "Not a table" when no rows were returned.

Slightly less trivial:

- Got nullable fields working.
- Got failed joins working.

### DIFF
--- a/modules/doobie/src/main/scala/doobiemapping.scala
+++ b/modules/doobie/src/main/scala/doobiemapping.scala
@@ -100,7 +100,7 @@ object DoobieMapping {
         else Join(child, parent)
       }
 
-      def toSql: String = s"${parent.toSql} = ${child.toSql}"
+      def toSql: String = s"LEFT JOIN ${child.table} ON ${parent.toSql} = ${child.toSql}"
     }
   }
 
@@ -226,13 +226,16 @@ object DoobieMapping {
         case _ => Fragment.empty
       }
 
-      val where = Fragments.whereAnd(preds ++ joins.map(join => Fragment.const0(join.toSql)): _*)
+      // we shouldn't need to do this because `tables` should really just be the root table
+      val tablesʹ = tables.filterNot(joins.map(_.child.table).toSet)
+
+      val where = Fragments.whereAnd(preds: _*)
 
       val select =
         Fragment.const0(
           s"""
           |SELECT ${cols.mkString(", ")}
-          |FROM ${tables.mkString(", ")}
+          |FROM ${tablesʹ.mkString(", ")}${if (joins.isEmpty) "" else joins.map(_.toSql).mkString("\n", "\n", "")}
           |""".stripMargin
         )
 
--- a/modules/doobie/src/main/scala/doobiemapping.scala
+++ b/modules/doobie/src/main/scala/doobiemapping.scala
@@ -197,39 +197,39 @@ object DoobieMapping {
       val preds = predicates.map {
         case (tpe, FieldEquals(fieldName, value)) =>
           val col = columnOfField(tpe, fieldName)
-          Fragment.const(s"${col.toSql} = ") ++ fr"$value"
+          Fragment.const(s"${col.toSql} =") ++ fr0"$value"
 
         case (tpe, FieldLike(fieldName, pattern, caseInsensitive)) =>
           val col = columnOfField(tpe, fieldName)
           val op = if(caseInsensitive) "ILIKE" else "LIKE"
-          Fragment.const(s"${col.toSql} $op ") ++ fr"$pattern"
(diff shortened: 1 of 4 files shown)


modules/doobie/src/main/scala/doobiemapping.scala
```diff
@@ -197,42 +197,45 @@ object DoobieMapping {
...
           val col = columnOfKey(tpe1, path.last)
-          Fragment.const(s"${col.toSql} = ") ++ fr"$value"
+          Fragment.const(s"${col.toSql} =") ++ fr0"$value"
 
         case _ => Fragment.empty
       }
 
-      val where = Fragments.whereAnd(preds ++ joins.map(join => Fragment.const(join.toSql)): _*)
+      // we shouldn't need to do this because `tables` should really just be the root table
+      val tablesʹ = tables.filterNot(joins.map(_.child.table).toSet)
```
tpolecat: I think `MappedQuery` should have `table: String` rather than `tables: List[String]` because any other tables we introduce will come from `Join`s ... ?

modules/doobie/src/test/scala/world/WorldSpec.scala
```diff
@@ -548,4 +574,65 @@ final class WorldSpec extends CatsSuite {
...
+  //   """
+
+  //   val compiledQuery = WorldQueryCompiler.compile(query).right.get
+  //   val res = WorldQueryInterpreter.fromTransactor(xa).run(compiledQuery).unsafeRunSync
+  //   //println(res)
+
+  //   assert(res == expected)
+  // }
+
+  test("no such country") {
```
tpolecat: This fails without the `| Nil` case above.

tpolecat:
The recursive queries can be implemented as joins rather than multiple fetches so but we need to implement table aliasing before we can do that. One thing at a time ;-)

tpolecat:
Ok nullable fields work now.

tpolecat:
Bit off more than I had intended but it seems to work.

milessabin:
Is this ready for review/merge or should I leave it until you get back?

modules/doobie/src/main/scala/doobiemapping.scala
```diff
@@ -298,21 +309,35 @@ case class Row(elems: List[Any]) {
...
 
-    val gets = cols.map(col => (typeToGet(col.tpe), NoNulls))
+  // Placeholder for nulls read from non-nullable columns introduced via an outer join.
+  case object FailedJoin
 
-    def unsafeGet(rs: ResultSet, n: Int): Row = {
-      Row(gets.zipWithIndex.map { case (g, i) => g._1.unsafeGetNonNullable(rs, n+i) })
+  def mkRead(cols: List[ColumnRef], isJoin: ColumnRef => Boolean): Read[Row] = {
+
+    def typeToGet(cr: ColumnRef): (ColumnRef, (Get[_], NullabilityKnown)) =
```
milessabin: I don't understand why we're wrapping the argument in a `ColumnRef` and returning a `ColumnRef` in an outer pair. Only the `Type` component of `cr` appears to be used in the body of `typeToGet` and only `_2` of the outer pair appears to be used in the sequel (ie. only the `(Type, NullabilityKnown)` elements of `gets` is used).

modules/doobie/src/test/scala/world/WorldSpec.scala
milessabin: :+1:

milessabin (commented):
I'm a little unsure how this interacts with the `StagingElaborator`. Previously it split DB queries into stages with deferred components, but presumably this is unnecessary now, at least for the cases we're considering here?

modules/doobie/src/main/scala/doobiemapping.scala
tpolecat: `unsafeGet` needs the `ColumnRef` to figure out whether the column is the product of an outer join or not. Is there a simpler way to figure this out?

milessabin: Right, but I'm still not quite seeing why the signature of `typeToGet` can't be,
```scala
def typeToGet(tpe: Type): (Get[_], NullabilityKnown)
```
I've reworked this area a fair bit, so it's not worth angsting over, but in case I'm missing something ...

milessabin:
I've reworked this in https://github.com/gemini-hlsw/gsp-graphql/pull/4/commits/35da44d9e64f6829bd9ed45477f095fdd1e2caf7 as part of #4.

PR #2, opened 2019-11-26
Cursor/DataTypeCursor improvements

milessabin:
* Added `tpe: Type` member to `Cursor`
* Thread GraphQL type though `DataTypeCursor`
* Parameterise `DataTypeCursor/Interpreter` with `PartialFunctions` instead of subclassing.

### DIFF
--- a/modules/core/src/main/scala/datatype.scala
+++ b/modules/core/src/main/scala/datatype.scala
@@ -7,66 +7,90 @@ import cats.Monad
 import cats.implicits._
 import io.circe.Json
 
+import Query.Wrap
 import QueryInterpreter.{ mkErrorResult, ProtoJson }
+import ScalarType._
 
-abstract class DataTypeQueryInterpreter[F[_]: Monad](schema: Schema)
-  extends QueryInterpreter[F](schema) {
-
-  def rootCursor(query: Query): Result[(Type, Cursor)]
+class DataTypeQueryInterpreter[F[_]: Monad](
+  schema: Schema,
+  root:   PartialFunction[String, (Type, Any)],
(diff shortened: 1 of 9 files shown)


modules/core/src/main/scala/datatype.scala
```diff
@@ -7,66 +7,90 @@ import cats.Monad
...
+  def asLeaf: Result[Json] = (tpe, focus) match {
+    case (StringType,  s: String)  => Json.fromString(s).rightIor
+    case (IntType,     i: Int)     => Json.fromInt(i).rightIor
+    case (FloatType,   d: Double)  => Json.fromDouble(d) match {
+        case Some(j) => j.rightIor
+        case None => mkErrorResult(s"Unrepresentable double %d")
+      }
+    case (BooleanType, b: Boolean) => Json.fromBoolean(b).rightIor
+    case (_: EnumType, e: Enumeration#Value) => Json.fromString(e.toString).rightIor
+    case _ => mkErrorResult(s"Expected Scalar type, found ${tpe.shortString}")
```
tpolecat: We can also get here in case of a mismatch (where the schema type and focus class type aren't compatible) so it might be good to include `focus` in the error message. Same comment throughout this class.

tpolecat (approved):
Nice improvement. End-user code is a lot cleaner. One comment on implementation error cases that may be obscured … feel free to improve or open a related issue and we can come back to it, either is fine.

modules/core/src/main/scala/datatype.scala
milessabin: Agreed.

PR #4, opened 2019-12-05
Schema composability improvements; integrated Robs nullability support

milessabin:
This is a fairly large reorganization to support the composition of schemas/models/interpreters more cleanly. Prior to this PR _component_ interpreters had to be parameterised with the _composed_ schema. Following this PR component interpreters are always parameterized with their local schemas whether they are part of a larger composition or being used standalone.

For example, suppose we have schema/model/interpreter A and schema/model/interpreter B. We could work with either individually, in which case the interpreter for A would be parameterized with the schema for A; and the interpreter for B would be parameterized with the schema for B. So far so good.

However if we wanted to compose the schemas/models we would not be able to simply glom the A and B interpreters together. Instead they would both need to be parametrized with the composed schema, despite their local behaviour and semantics being unchanged. This opened up many opportunities for error, mainly around confusions between types belonging to the local and the composed schemas, and types leaking between components inappropriately.

Following this PR, component interpreters always work with their corresponding component types. The elaboration process during compilation is now responsible for ensuring that when a part of a query is dispatched to a component interpreter it's given an expected type from drawn from the corresponding component schema.

This had a fairly significant impact on the DB mapping (a parameter of the DB-backed interpreter). Previously it had to be defined in terms of a specific schema, which was problematic where there was more than one schema in play (ie. component vs. composed, as above). It turned out that the field types mentioned in the mapping were redundant (ie. could be inferred from the mapping and the actual schema it was used in conjunction with) and could be removed, simplifying the definition of the mapping with no loss of safety.

This left a small number of hard references to top-level types from a fixed schema in the mapping which would need to be rebound somehow if moving between schemas. In the end I decided to bind these by name rather than by reference to a type (ie. a Scala reference to a value of type `Type`). I've managed to convince myself that while this might seem like a reduction in safety, it actually isn't. There's work still to be done here, checking that a mapping is consistent with the schema it's used with, but that's work that would have had to be done anyway.

Relatedly I cleaned up the distinction between _fields_ (things which appear in the GraphQL schema) and _attributes_ (things which appear only in the model, eg. ids for joins) and between both of these and keys. In the DB mapping, attributes are assigned an arbitrary Scala type (rather than a GraphQL type) so long as there is a Doobie `Get` instance available for it.

There were lots of conflicts with @tpolecat's [PR](https://github.com/gemini-hlsw/gsp-graphql/pull/1) adding nullability support, so I've also reworked that here in the final commit.

### DIFF
--- a/modules/core/src/main/scala/compiler.scala
+++ b/modules/core/src/main/scala/compiler.scala
@@ -87,7 +87,7 @@ object QueryCompiler {
             }
           }
 
-        case w@Wrap(fieldName, child) => apply(child, tpe.underlyingField(fieldName)).map(ec => w.copy(child = ec))
+        case w@Wrap(_, child)         => apply(child, tpe).map(ec => w.copy(child = ec))
         case g@Group(queries)         => queries.traverse(q => apply(q, tpe)).map(eqs => g.copy(queries = eqs))
         case u@Unique(_, child)       => apply(child, tpe.nonNull).map(ec => u.copy(child = ec))
         case f@Filter(_, child)       => apply(child, tpe.item).map(ec => f.copy(child = ec))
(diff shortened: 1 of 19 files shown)


tpolecat (approved):
These look like good improvements. I'm pretty confident my test cases cover the important cases for nulls and outer joins so :+1:.

PR #6, opened 2020-01-07
Support fragments

milessabin:
* Added `Narrow` operator to the query algebra. When evaluated it will refine the cursor to the target type if possible and its child will be evaluated relative to that. If it can't be so refined the child is ignored.
* Added a `<:<` (subtype) operator to `Type`.
* Added `narrowsTo` and `narrow` methods to `Cursor` for refining runtime types. `DataTypeCursor` implements these in terms of an optional `PartialFunction` from the focus and the target type to the refined focus value.
* Compiler now supports queries containing inline and standalone fragments. Fragment spreads are now translated to `Narrow` nodes.
* Request document parsing now takes an optional operation name argument which if present will be used to select which operation to perform from a multi-operation request.
* Phase `apply` methods now take a `Schema` argument as well as a target type.

### DIFF
--- a/demo/src/main/scala/starwars/StarWarsService.scala
+++ b/demo/src/main/scala/starwars/StarWarsService.scala
@@ -15,7 +15,7 @@ import edu.gemini.grackle.QueryInterpreter
 
 // #service
 trait StarWarsService[F[_]]{
-  def runQuery(op: String, query: String): F[Json]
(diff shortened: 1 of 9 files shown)


tpolecat (approved):
:+1: This looks good. I will play with the latest version this week so we have time for an informed chat before you finish up. Thanks!

PR #7, opened 2020-01-10
Add support for GraphQL introspection

milessabin:
This PR adds a complete implementation of GraphQL introspection and some preliminary integration of that in the demo project.

The demo will now respond correctly to GraphQL Playground introspection queries which means that the playground "Schema" and "Docs" tabs are populated with information from the actual demo schema rather than the canned (and not completely accurate) version which was provided previously.

A current limitation is that introspection queries currently have to be given an explicit operation name of `IntrospectionQuery` and hence that, for queries of this sort, shorthand notation can't be used. Concretely this means that the following query in the playground will work as expected,

```
query IntrospectionQuery {
  __schema {
    types {
      name
    }
  }
}
```
yielding,
```json
{
  "data": {
    "__schema": {
      "types": [
        {
          "name": "Query"
        },
        {
          "name": "Episode"
        },
        {
          "name": "Character"
        },
        {
          "name": "Human"
        },
        {
          "name": "Droid"
        }
      ]
    }
  }
}
```
but the shorthand form,
```
query {
  __schema {
    types {
      name
    }
  }
}
```
will not.


tpolecat (approved):
Very nice!

PR #9, opened 2020-01-10
make Phase a monoid, rename to Elaborator

tpolecat:
This occurred to me but I don't know if it's sensible or not. `QueryCompiler` is also a monoid with "or else" composition but I'm not sure it's useful because you need to know which interpreter to pass the result to.

### DIFF
--- a/demo/src/main/scala/starwars/StarWarsData.scala
+++ b/demo/src/main/scala/starwars/StarWarsData.scala
@@ -136,7 +136,7 @@ object StarWarsData {
 
 object StarWarsQueryCompiler extends QueryCompiler(StarWarsSchema) {
   // #elaborator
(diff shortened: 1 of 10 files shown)


milessabin:
Minor point first: although it's the case at the moment, a phase needn't necessarily be an elaboration, eg. we're probably going to need one or more validation phases which won't involve elaborating (ie. transforming) the query. So :-1: to the renaming.

I'm not quite seeing the `orElse` aspect here, at least not with respect to phases. If _any_ of the phases fail then the overall query evaluation fails at that point.

In general composing interpreters involves composing corresponding phases. Eg. to properly compose the introspection mechanism with a vanilla model we need to combine the `SelectElaborator`s of each and have the composite run before any subsequent phase. Which is to say that there's more structure in phases than is just captured by a monoid.

Issue #13, opened 2020-01-13
Add support for multiple root queries


milessabin:
See discussion [here](https://github.com/graphql/graphql-spec/issues/29). Depends on #12.

This might also interact with staged query coalescing.

milessabin:
Fixed in #39.

PR #39, opened 2020-01-17
Add support for field aliases and multiple root fields

### LINKED ISSUE #12: Add support for field aliases
See [2.7](https://graphql.github.io/graphql-spec/June2018/#sec-Field-Alias).

milessabin:
It is now possible for a query to contain more than one root field as described in the spec discussion linked to in #13. Where the root queries differ only in their arguments they must be renamed using
field aliases (described in section 2.7 of the spec) to ensure that their result fields don't conflict.

Adding the support for renaming revealed a bug in the schema/interpreter composition mechanism which is now fixed.

Adding support for multiple root nodes revealed a bug in the handling of `Unique` queries where the expected type is nullable. Prior to this commit, if there was no unique match the query would fail incorrectly rather than returning a null value. This is now also fixed.

Fixes #12 and #13.

### DIFF
--- a/modules/core/src/main/scala/compiler.scala
+++ b/modules/core/src/main/scala/compiler.scala
@@ -70,14 +70,18 @@ object QueryParser {
     }
 
   def parseSelection(sel: Selection, typeCondition: Option[String], fragments: Map[String, FragmentDefinition]): Result[Query] = sel match {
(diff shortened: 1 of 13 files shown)


tpolecat (approved):
This looks good. No real increase in complexity anywhere, which is good validation. 👍

PR #40, opened 2020-01-24
Added initial support for query variables

milessabin:
* The compiler now takes an additional argument representing query variables represented as Json.
* Bindings refactored, pushing the type into the value in preparation for supporting input lists and objects.


tpolecat (approved):
👍 looks great!

I continue to be pleased that chipping away at the spec is resulting in nicely incremental changes.

PR #43, opened 2020-02-03
Fully integrate introspection

### LINKED ISSUE #34: Complete introspection integration
Currently the introspection implementation operates alongside the non-introspective mechanism distinguished by an explicit operation name `IntrospectionQuery`. It should be properly composed with the non-introspection mechanism, ie. introspection queries should be available as-if defined as part of the non-introspective schema.

milessabin:
* Introspection queries are now supported by all interpreters and standalone introspection interpreter has been dropped.
* Added support for `__typename`.
* Fixed union type issues revealed in interactions with introspection.
* Adapted the demo service interface for integrated introspection and brought GET query handling up to par with POST.

Fixes #34.


tpolecat (approved):
Cool, this looks good.

PR #44, opened 2020-02-04
Added support for skip/include directives

### LINKED ISSUE #18: Add support for @skip and @include directives
See [3.13.1](https://graphql.github.io/graphql-spec/June2018/#sec--skip) and [3.13.2](https://graphql.github.io/graphql-spec/June2018/#sec--include).

It should be possible to eliminate these during elaboration.

Depends on #17.

milessabin:
Fixes #18.

### DIFF
--- a/modules/core/src/main/scala/parser.scala
+++ b/modules/core/src/main/scala/parser.scala
@@ -137,7 +137,7 @@ object Parser {
     (Name <~ keyword(":")) ~ Value
 
   lazy val VariableDefinitions =
-    parens(many(VariableDefinition)).token
+    parens(many(VariableDefinition <~ opt(keyword(",")))).token
 
   lazy val VariableDefinition: Parser[Ast.VariableDefinition] =
     for {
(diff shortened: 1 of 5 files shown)


modules/core/src/main/scala/parser.scala
```diff
@@ -137,7 +137,7 @@ object Parser {
     (Name <~ keyword(":")) ~ Value
 
   lazy val VariableDefinitions =
-    parens(many(VariableDefinition)).token
+    parens(many(VariableDefinition <~ opt(keyword(",")))).token
```
tpolecat: My understanding is that commas are whitespace in graphql so I think we need to solve this more generally. But this is ok for now.

tpolecat (approved):

modules/core/src/main/scala/parser.scala
milessabin: That's right. I need to add support for comments to enable schema (de) serialisation and I'll pick this up in the process.

PR #47, opened 2020-03-26
Added simple shapeless-based derivation for ADT backed models

milessabin:
A fairly simple minded `Cursor` derivation scheme for ADT-backed models.

This PR adds a type class `CursorBuilder[T]` which constructs `Cursors` for values of Scala type `T`. Instances are provides for primitive types, `List`, `Option` and Scala enumerations.

Circe-style semiauto derivation is provided for,
* sealed traits, which correspond to GraphQL interface types
* case classes, which correspond to GraphQL object types
* Any type with a Circe `Encoder` instance, which corresponds to a (custom) GraphQL scalar type.

Usage is similar to Circe,

```scala
import edu.gemini.grackle.generic.semiauto._

sealed trait Character {
  def id: String
  def name: Option[String]
  def appearsIn: Option[List[Episode.Value]]
  def friends: Option[List[String]]
}

object Character {
  implicit val cursorBuilder: CursorBuilder[Character] =
    deriveInterfaceCursorBuilder[Character]
}

case class Human(
  id: String,
  name: Option[String],
  appearsIn: Option[List[Episode.Value]],
  friends: Option[List[String]],
  homePlanet: Option[String]
) extends Character

object Human {
  implicit val cursorBuilder: CursorBuilder[Human] =
    deriveObjectCursorBuilder[Character]
}

object CustomScalars {
  implicit val zdtCursorBuilder: CursorBuilder[ZonedDateTime] =
    deriveLeafCursorBuilder[ZonedDateTime]
}
```

Derived `ObjectCursorBuilders` may be transformed,
* to map their field names
* to transform field values

The latter is particularly useful in situations where the GraphQL schema is naturally recursive, but where the Scala model ties the knot indirectly, eg. via an id. In this case the id can be mapped to the specified value by the cursor, allowing query navigation through the unfolding structure.

```scala
object Human {
  implicit val cursorBuilder: CursorBuilder[Human] =
    deriveObjectCursorBuilder[Human]
      .transformField("friends")(_.friends match {
        case None => None.rightIor
        case Some(ids) =>
          ids.traverse(id => characters.find(_.id == id)
            .toRightIor(mkOneError(s"Bad id '$id'"))).map(_.some)
      })
}
```

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -106,7 +106,7 @@ lazy val generic = project
 lazy val demo = project
   .in(file("demo"))
   .enablePlugins(AutomateHeaderPlugin)
-  .dependsOn(core, doobie)
+  .dependsOn(core, doobie, generic)
   .settings(commonSettings)
   .settings(
     name := "gsp-graphql-demo",
--- a/build.sbt
+++ b/build.sbt
@@ -10,6 +10,7 @@ val jawnVersion                 = "0.14.3"
 val kindProjectorVersion        = "0.10.3"
 val logbackVersion              = "1.2.3"
 val log4catsVersion             = "1.0.1"
+val shapelessVersion            = "2.3.3"
 val slf4jVersion                = "1.7.29"
 
 inThisBuild(Seq(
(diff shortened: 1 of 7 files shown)


tpolecat (approved):
Cool! :+1:

PR #50, opened 2020-04-01
Allow components to construct lists as well as subobjects

milessabin:
Prior to the PR it was only possible to delegate the construction of a field value to a component. As of this PR it is now possible to delegate the construction of list elements to a component.

### DIFF
--- a/modules/core/src/main/scala/query.scala
+++ b/modules/core/src/main/scala/query.scala
@@ -48,7 +48,12 @@ object Query {
 
   /** A Group of sibling queries at the same level */
   case class Group(queries: List[Query]) extends Query {
-    def render = queries.map(_.render).mkString(", ")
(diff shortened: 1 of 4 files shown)


modules/core/src/main/scala/query.scala
```diff
@@ -608,6 +628,20 @@ object QueryInterpreter {
       else
         wrap(ProtoArray(elems))
 
+    /**
+     * Select a value from a possibly partial object.
+     *
+     * If the object is complete the selection will be a coplete
```
tpolecat, line 634: *complete

tpolecat (approved):
:+1:

modules/core/src/main/scala/query.scala
milessabin: Well spotted!

Issue #19, opened 2020-01-13
Add support for @deprecated directive


milessabin:
See [3.13.3](https://graphql.github.io/graphql-spec/June2018/#sec--deprecated).

keirlawson:
Is this now resolved or is there work outstanding?

milessabin:
It'd be worth checking through the spec but, yes, I think so.

milessabin:
Fixed a very long time ago in #53.

PR #53, opened 2020-05-16
add support for deprecated directive in schema

jbwheatley:
see #19 
confirms to http://spec.graphql.org/June2018/#sec--deprecated

### DIFF
--- a/modules/core/src/main/scala/schema.scala
+++ b/modules/core/src/main/scala/schema.scala
@@ -675,7 +675,7 @@ case class Directive(
  * GraphQL schema parser
  */
 object SchemaParser {
-  import Ast.{ Type => _, Value => _, Directive => _, _ }, OperationType._
+  import Ast.{ Type => _, Value => _, Directive => DefinedDirective, _ }, OperationType._
 
   /**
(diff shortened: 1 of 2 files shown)


milessabin:
This is more great stuff ... thanks!

I think the only thing missing is a test that deprecation is visible via schema introspection. Would you like to take a swing at adding a test along those lines?

jbwheatley:
Yep, I can do that!

jbwheatley:
@milessabin looks like there's already some testing around deprecation in introspection https://github.com/gemini-hlsw/gsp-graphql/blob/63c5ec1dfc6e1bab079f6c4b85b67e8319c56f30/modules/core/src/test/scala/introspection/IntrospectionSpec.scala#L704

milessabin:
@jbwheatley Yup, but that's vai a manually constructed schema value with a deprecated. It'd be nice to have something that started from SDL as well. Or maybe instead? Perhaps replace the manually constucted schemas in those tests with SDL?

milessabin:
But, actually, that could just as well be a follow on PR ... I think this is good to go as it is :smile:

PR #54, opened 2020-05-19
Miscellaneous fixes and enhancements

milessabin:
+ Allow comma separated selector arguments
+ Improved error message for queries with no operations
+ Map ID type to SQL string in Doobie mapping
+ Don't parse Scalar types as `TypeRefs`
+ Bumped Doobie version
+ Tidied up logging


tpolecat (approved):
:+1:

PR #57, opened 2020-05-26
Use TestContainers

milessabin:
Switched to [Testcontainers-scala](https://github.com/testcontainers/testcontainers-scala) instead of manual `docker-compose up`.


tpolecat (approved):
Huh, neato. :+1:

Issue #68, opened 2020-08-14
Validate schema


milessabin:
Currently there's no validation of schemas beyond basic syntactic well-formedness. Consequently it's possible to construct schemas with references to undefined types (typically due to a typo in the definition or use) or multiple definitions (eg. I've accidentally defined a first as a scalar then as an enum, but forgotten to remove the first definition). Invalid schemas of this sort can result in all sorts of confusing runtime failures later.

The GraphQL spec has a fairly comprehensive list of validation criteria for _queries_. The validation criteria for schemas is less explicit and spread throughout section 3 of the spec: https://spec.graphql.org/June2018/#sec-Type-System. In practice I think that checking for references to undefined or multiply defined types will cover the most common issues.

GraphQL schemas are represented as values of the `Schema` types and the validation above would be a matter of checking the `NamedType` elements of its `types` member for consistency and completeness.

yasuba:
I've agreed with Miles to take a look at this issue.

milessabin:
Mostly done in #68.

There are a few additional things we could add. For instance checking that a object type defines all the members specified by the interfaces it implements.

keirlawson:
Was wondering what is left to do here? Happy to have a go at contributing further validation if there is more required.

milessabin:
I think probably needs a quick review of the spec before closing.

PR #78, opened 2020-08-24
issue 68 - validate schema

yasuba:
Having published locally and tested against a service with a misspelling in the schema, the error surfaces at runtime, not compile time. Although it does say `"message" : "Reference to undefined type: Episod"`. I shouldn't think this the desired behaviour.


tpolecat:
Does this fully address #68? If so we can close it.

milessabin:
:+1: from me too.

There are a few additional things we could add. For instance checking that a object type define all the members specified by the interfaces it implements.

PR #81, opened 2020-09-14
Compiler parses into UntypedOperation (Untyped[Query|Mutation|Subscription])

rpiaggio:
* Introduced `UntypedOperation`, which can be `UntypedQuery`, `UntypedMutation` or `UntypedSubscription` and contains a `Query` and `UntypedVarDefs`.
* `QueryParser`'s `parse*` methods that returned tuples now return an `UntypedOperation`.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -43,9 +43,13 @@ bin/
 # Mill
 out/
 
+# VSCode
+.vscode/
+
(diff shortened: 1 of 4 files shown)


milessabin (approved):
LGTM! Thanks :-)

Issue #73, opened 2020-08-14
Add support for Postgres jsonb columns


milessabin:
Following on from #72 it should be possible to add support for Postgres jsonb columns similarly to how array columns where added in #66.

milessabin:
Fixed in #82.

PR #82, opened 2020-09-18
Added support for Circe backed models and Postgres jsonb in Doobie backed models

### LINKED ISSUE #72: Add support for circe backed models
Currently Grackle has support for models back by Scala ADTs (via either manual or derived mappings) and SQL databases via doobie.

It would be nice to also be able to use a (collection of) circe Json objects as the model with GraphQL queries having their natural interpretation as traversals through and selections from those objects.

A basic implementation should be fairly straightforward ... we would need a new subtype of `Mapping` which provided a Grackle `Cursor` backed by a circe cursor.

milessabin:
Fixes #72 and #73.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -89,7 +89,7 @@ lazy val doobie = project
   .in(file("modules/doobie"))
   .enablePlugins(AutomateHeaderPlugin)
   .disablePlugins(RevolverPlugin)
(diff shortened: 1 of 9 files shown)


tpolecat (approved):
:+1: this is cool

PR #84, opened 2020-09-30
Mappings validation

yasuba:


modules/core/src/main/scala/mapping.scala
```diff
@@ -17,6 +17,26 @@ trait Mapping[F[_]] {
...
+    def typeMappingFieldNames: List[String] = typeMappings.collect {
+      case om: ObjectMapping => om.fieldMappings.flatMap {
+        case _: CursorField[_] => Nil
+        case _: CursorAttribute[_] => Nil
+        case fm => List(fm.fieldName)
+      }
+    }.flatten
+
+    val mappingTypesExistInSchema: Boolean =
+      typeMappingNames.forall(name => schema.types.map(_.name).contains(name))
```
yasuba, line 32: It would be good to have some kind of logging or feedback about which types and fields are missing.

milessabin: Maybe have `validate` return a list of errors encoded as json? See `QueryInterpreter.mkOneError`.

milessabin (commented):
Just that one comment, and needs rebasing, but otherwise LGTM!

modules/core/src/main/scala/mapping.scala
yasuba: Sounds good to me.

modules/core/src/main/scala/mapping.scala
```diff
@@ -5,21 +5,43 @@ package edu.gemini.grackle
...
   implicit val M: Monad[F]
 
   val schema: Schema
   val typeMappings: List[TypeMapping]
-
+  
   def typeMapping(tpe: Type): List[TypeMapping] =
     typeMappings.filter(_.tpe.nominal_=:=(tpe))
 
+  def validate: Either[NonEmptyChain[Json], String] = {
```
milessabin, line 23: Ahh ... right, I see what you means about the result type looking odd. Why not return `Chain[Json]` where an empty chain indicates no errors detected?

milessabin (commented):
Aside from the signature tweak, LGTM :-)

modules/core/src/main/scala/mapping.scala
yasuba: Ok, yes that sounds good.

milessabin (approved):
LGTM!

PR #87, opened 2020-10-20
Ensure required columns for computed fields/attributes

milessabin:
Adds a required fields/attributes member to `CursorField` and `CursorAttribute` allowing the Doobie mapping to ensure that columns which are not otherwise referenced are fetched and available for the evaluation of the computed field or attribute.


tpolecat (approved):
Excellent. 👍

PR #88, opened 2020-10-20
fix ci to fetch tags

tpolecat:
`actions/checkout@v2` does not fetch tags, but it will if you set `fetch-depth` to zero. Snapshot versions should be correct now, fingers crossed.


milessabin (approved):
LGTM!

PR #89, opened 2020-10-23
Mapping validation additions

milessabin:
+ CursorFields/Attributes now have a list of required mappings so we can check for their presence.
+ We shouldn't report attributes as missing fields.


yasuba (approved):
lgtm

PR #94, opened 2020-11-17
Added Project to push a predicate into a subtree

milessabin:
WIP. This will need fleshing out in terms of SQL subqueries ... see the two disabled tests in `ProjectionSpec`.


tpolecat (approved):
:+1: (pls add an issue for the ignored tests)

PR #96, opened 2020-11-17
detect enum value duplicates

### LINKED ISSUE #92: Schema permits duplicate enum values
If I'm reading the spec correctly (and understanding the code correctly!) duplicate enum values in schemas should not be permitted (https://spec.graphql.org/June2018/#sec-Enums), but presently the following test fails:

```scala
val schema = Schema(
      """
         enum Direction {
          NORTH
          NORTH
        }
    """
    )

    assert(schema.isLeft)
```

Seems to me that we should be holding these values as a `Set` rather than a `List`? Happy to work on a fix if someone can clarify whether or not this is intended behaviour :)

keirlawson:
Fixes #92

Feedback appreciated :)

### DIFF
--- a/modules/core/src/main/scala/schema.scala
+++ b/modules/core/src/main/scala/schema.scala
@@ -4,14 +4,14 @@
 package edu.gemini.grackle
 
 import atto.Atto._
-import cats.data.Ior
+import cats.data.{Ior, NonEmptyChain}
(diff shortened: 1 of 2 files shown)


milessabin (approved):
This is spot on ... many thanks!

PR #97, opened 2020-11-20
Followup to the skunk refactoring

milessabin:
+ Eliminated `SqlLeafMapping` now that DB codecs are carried by `ColumnRefs`.
+ Added `Projection` and tests which landed after the skunk refactoring. Skunk tests are completely disabled at the moment.
+ Swapped out old for new doobie mapping


tpolecat (approved):
🎉

PR #98, opened 2020-11-25
Validate interface implementation

keirlawson:
Validates that an object correctly implements any specified interfaces per #68.

Does not include checking that an interface correctly implements another, nor that an object implements all transitive interfaces.

Feedback appreciated :)

### DIFF
--- a/modules/core/src/main/scala/schema.scala
+++ b/modules/core/src/main/scala/schema.scala
@@ -955,11 +955,11 @@ object SchemaValidator {
   }
 
   def validateImpls(definitions: List[TypeDefinition]): List[Json] = {
-    val interfaces = definitions.collect[InterfaceTypeDefinition] {
+    val interfaces = definitions.collect {
       case a: InterfaceTypeDefinition => a
     }
 
-    val objects = definitions.collect[ObjectTypeDefinition] {
+    val objects = definitions.collect {
       case a: ObjectTypeDefinition => a
     }
 
--- a/modules/core/src/main/scala/schema.scala
+++ b/modules/core/src/main/scala/schema.scala
@@ -5,14 +5,14 @@ package edu.gemini.grackle
 
 import atto.Atto._
 import cats.data.{Ior, NonEmptyChain}
-import cats.implicits._
 import io.circe.Json
-import Ast.{EnumTypeDefinition, ObjectTypeDefinition, TypeDefinition}
+import Ast.{EnumTypeDefinition, FieldDefinition, InterfaceTypeDefinition, ObjectTypeDefinition, TypeDefinition}
 import QueryInterpreter.{mkError, mkErrorResult, mkOneError}
 import ScalarType._
 import Value._
 import cats.data.Ior.Both
 import cats.kernel.Semigroup
+import cats.implicits._
 
 /**
  * Representation of a GraphQL schema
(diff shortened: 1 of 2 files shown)


modules/core/src/main/scala/schema.scala
```diff
@@ -950,11 +950,45 @@ object SchemaValidator {
   def validateSchema(namedTypes: Result[List[NamedType]], defns: List[TypeDefinition]): Result[List[NamedType]] = {
     val undefinedResults = checkForUndefined(checkForDuplicates(namedTypes), defns)
 
-    val duplicates = NonEmptyChain.fromSeq(checkForEnumValueDuplicates(defns))
+    val errors = NonEmptyChain.fromSeq(checkForEnumValueDuplicates(defns) ++ validateImpls(defns))
+    errors.map(addLeft(undefinedResults, _)).getOrElse(undefinedResults)
+  }
+
+  def validateImpls(definitions: List[TypeDefinition]): List[Json] = {
+    val interfaces = definitions.collect[InterfaceTypeDefinition] {
```
tpolecat, line 958: (The type arguments here and below are unnecessary.)

modules/core/src/main/scala/schema.scala
```diff
@@ -950,11 +950,45 @@ object SchemaValidator {
...
+          case None => List(mkError(s"Interface ${ifaceName.astName.value} implemented by object ${obj.name.value} is not defined"))
+        }
+      }
+    }
+  }
+
+  def checkImplementation(obj: ObjectTypeDefinition, interface: InterfaceTypeDefinition): List[Json] = {
+    interface.fields.flatMap { ifaceField =>
+      obj.fields.find(_.name == ifaceField.name).map { matching =>
+        if (ifaceField.tpe != matching.tpe) {
```
tpolecat, line 979: Should these be compared with `=:=`?

tpolecat (commented):
This looks good, thanks! I'll wait for @milessabin to give the final approval.

modules/core/src/main/scala/schema.scala
keirlawson: Do you mean `=!=`?

keirlawson: So none of these types have an `Eq` instance at present so far as I can see? Happy to implement one if you reckon this is the way to go

tpolecat: Sorry, `Type` specifically has an `=:=` operator that's like `==` but compares dealiases the types before comparing them. I don't know if it matters here. @milessabin will.

milessabin: Yes, you should compare types with `=:=` in the typical case.

milessabin:
@keirlawson I'm _really_ sorry it's taken me so long to get to this ... things were a bit hectic in the run up to the end of the year.

Aside from the `==` vs. `=:=` issue this look great :-)

modules/core/src/main/scala/schema.scala
keirlawson: These are `Ast.Type`s though, not `Schema.Type`s, so there doesn't seem to be `=:=` defined unless I'm missing something?

milessabin: You're absolutely right! Yes, this is correct just as it is.

milessabin (approved):
LGTM! Many thanks!

PR #99, opened 2020-12-21
SQL staging fixes

milessabin:
Some fixes for grouping across stages where one or more groups might be empty or where we join through an object but don't explicitly refer to any of its fields.


cquiroz (approved):
LGTM

tpolecat (approved):

PR #101, opened 2021-01-05
Bumped versions; fixed lint errors

milessabin:
Bumped all the versions and fixed the resulting glitches.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -50,6 +50,7 @@ out/
 .bloop/
 .metals/
 metals.sbt
+.bsp
 
 # Vim
 *.swp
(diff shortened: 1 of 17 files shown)


milessabin:
Hmm ... that semanticdb dependency failure is odd. It looks like it should have been fixed in SBT 1.4.4.

tpolecat:
I think we need to update `sbt-lucuma`, which is pulling in an old version of scalafix. For now try adding scalafix expliictly with the latest version.

milessabin:
That appears to have done the trick ... thanks!

tpolecat:
> Is there any reason why we can't point Scala Steward at Grackle?

no, we should do that. and set up mergify as well. feel free, or open an issue and one of us can do it

tpolecat (approved):

PR #103, opened 2021-01-11
implementing unionTypes in sql

yasuba:


milessabin:
This looks great. Have you tried the tests against the skunk backend as well?

tpolecat:
Yeah the spec need to be mirrored over in `skunk/` (this is why all the `SqlWoozleSpec`s are factored out). Let me know if you need help doing this.

tpolecat (approved):
I'm impressed that this was such a small change! :+1:

PR #106, opened 2021-01-14
Don't ignore invalid document tail.

tpolecat:
There was an issue with the parser where it would never actually fail, but instead would stop when it reached an invalid definition and yield the [possibly empty] valid prefix. This fixes it to require that the entire document be valid.


milessabin (approved):
LGTM!

PR #107, opened 2021-01-15
improved mapping validation

tpolecat:
This adds a few more mapping validation checks, and moves it off to the side where it's a bit easier to extend with new checks as we discover them. The largest design change is in the intended usage: instead of yielding Json to be returned to clients, it yields a chain of `MappingValidator.Failure` which are structured values that can be inspected or converted to messages for the programmer, which are formatted and include source positions for mapping constructions, for example:

image

The TL;DR is that programmers can add `mapping.unsafeValidate()` in order to surface problems during development.

Note that many kind of mappings are opaque and cannot be validated. These provoke low-severity warnings that are elided by default; you can pass `Severity.Info` to any of the `validate` methods if you wish to see these as well.

API changes to the core model are mostly to provide source positions (and in some case type names) to allow for better error reporting.


milessabin (approved):
This looks awesome!

PR #108, opened 2021-01-19
Include required fields/mappings for introspection only selections

milessabin:
Previously if a sub-selection only contained introspection fields, the required fields and mappings wouldn't be included in the mapped query leading to a crash a runtime.


tpolecat (approved):
:+1:

yasuba (approved):

PR #111, opened 2021-01-19
fix interpretation of schema definition, select correct root type

tpolecat:
This changes `QueryCompiler.compile` to return `Result[Operation]` where `Operation(query: Query, rootTpe: Type)` which includes the selected root type. `QueryExecutor` uses this by default when invoking the interpreter (it had been using `schema.queryType` in all cases). 

`Operation` is a bit overloaded so maybe `RootedQuery` or something would be a better name.

This fixes a few problems:

- It was not possible to declare `schema { ... }` because it was expecting `NamedType` fields but they're actually `NullableTyped(NamedType)` as parsed.
- In the absence of a `schema { ... }` declaration we were defaulting to `schema { query: Query }` which ignored `type Mutation` and `type Subscription` when present. We now delegate to the [correct] superclass behavior in this case.

(this supersedes #110)

### DIFF
--- a/modules/core/src/main/scala/compiler.scala
+++ b/modules/core/src/main/scala/compiler.scala
@@ -63,7 +63,7 @@ object QueryParser {
       val vs = vds.map {
         case VariableDefinition(nme, tpe, _) => UntypedVarDef(nme.value, tpe, None)
       }
-      q.map(q => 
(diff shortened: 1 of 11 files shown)


milessabin (approved):
I'd missed that! LGTM ...

PR #112, opened 2021-01-20
pull out predicate and queryinterpreter

tpolecat:
This pulls `Predicate` and `QueryInterpreter` out into their own compilation units. No other changes.


milessabin (approved):
LTGM.

PR #113, opened 2021-01-20
Validation fixes

milessabin:
+ Attribute mappings are no longer rejected as unknown fields.
+ Enum and list types are accepted as leaf mappings.
+ Input object types no longer require mappings.
+ Introduced PrimitiveMapping as a checkable primitive mapping which does not require codecs.


tpolecat (approved):
Good catches, all. Thanks! 👍🏻

PR #115, opened 2021-02-05
Added scoped environment to Cursor

milessabin:
* Added environment map, lookup, and parent cursor to cursors.
* Added `Environment` element to query algebra.
* Previous `Env` is now `Vars`.

See [here](https://github.com/gemini-hlsw/gsp-graphql/blob/106c8f238699ff40c5cfc18051c0e738b9d6e751/modules/core/src/test/scala/compiler/Environment.scala) for a full example. We can add to the environment of a cursor in a `SelectElaborator`, eg.,

```scala
case Select("sum", List(Binding("x", IntValue(x)), Binding("y", IntValue(y))), child) =>
  Environment(Env("x" -> x, "y" -> y), Select("sum", Nil, child)).rightIor
```

and then retrieve these values via a `Cursor`, eg. in a `CursorField`,

```scala
ObjectMapping(
  tpe = NestedType,
  fieldMappings = List(CursorField("sum", sum))
)

def sum(c: Cursor): Result[Int] =
  (for {
    x <- c.env[Int]("x")
    y <- c.env[Int]("y")
  } yield x+y).toRightIor(mkOneError(s"Missing argument"))
```

### DIFF
--- a/modules/circe/src/main/scala/circemapping.scala
+++ b/modules/circe/src/main/scala/circemapping.scala
@@ -8,6 +8,7 @@ import cats.Monad
 import cats.implicits._
 import io.circe.Json
 
+import Cursor.Env
 import QueryInterpreter.{mkErrorResult, mkOneError}
 import ScalarType._
 import org.tpolecat.sourcepos.SourcePos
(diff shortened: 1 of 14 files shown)


yasuba (approved):

tpolecat (approved):
This looks good. It's better than what I had in mind for this case. :+1:

PR #119, opened 2021-02-18
Ensure there is at most one join on any given child table

milessabin:
Currently the SQL compiler doesn't generate subqueries in where clauses, and so projected predicates (which would naturally compile to subqueries in where clauses) end up being compiled as joins. Where there is more than one of these, or where one of these is combined with a legitimate join, we end up generating multiple left joins with the same child table. Postgres rejects this.

This PR is a workaround for that, in advance of a reworking of the compiler which generates subqueries, which checks for duplicate joins and converts all but the first to where clauses. I freely admit this is a hack.

### DIFF
--- a/modules/sql/src/main/scala/SqlMapping.scala
+++ b/modules/sql/src/main/scala/SqlMapping.scala
@@ -658,6 +658,20 @@ trait SqlMapping[F[_]] extends CirceMapping[F] with SqlModule[F] { self =>
         orderJoins(Set(rootTable), joins, Nil).reverse
       }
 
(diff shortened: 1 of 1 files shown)


yasuba (approved):

tpolecat (changes requested):
This rewrite is not valid in general so Miles is going to add some comments explaining what's going on, and will open a related issue to fix this (we discussed offline).

tpolecat (approved):

milessabin:
Umbrella issue here: https://github.com/gemini-hlsw/gsp-graphql/issues/120.

PR #124, opened 2021-03-01
Validate field arguments and input object fields

milessabin:
Prior to this commit field argument and input object fields which were present in the schema would be validated against the schema as expected. However, field arguments and input value fields which were not mentioned in the schema would be silently dropped.

These bogus arguments and fields are now reported.


tpolecat (approved):
👍🏻

PR #127, opened 2021-03-25
WIP: mutation

tpolecat:
This adds support for mutation as follows:

- `RootMapping` now has a `mutation` member of type `Mutation` (a wrapper for `(Query, Env) => F[Result[(Query, Env)]]`). All constructors default this value to `Mutation.None`, which does nothing.
- A `Mutation` performs effects in `F` and yields a [possibly updated] query and env.
- The `QueryInterpreter` executes the mutation and uses the [possibly updated] query and env to construct the result cursor.
- For the doobie and Skunk mappings the updates and subsequent queries run in separate transactions, which I think is actually what we want, given the "mutate and then query" semantics in the spec. In any case _not_ doing it this way would be a significant complication so I'm calling it good for now.

Todo:

- [x] Add an database insert test that generates a new key.
- [x] Implement database tests for doobie.

It's a pretty small change, which makes me hopeful. More comments inline.

### DIFF
--- a/modules/core/src/main/scala/mapping.scala
+++ b/modules/core/src/main/scala/mapping.scala
@@ -4,7 +4,7 @@
 package edu.gemini.grackle
 
 import cats.Monad
-import cats.data.Ior
+import cats.data.{ Ior, IorT }
 import cats.implicits._
 import io.circe.{Encoder, Json}
 import org.tpolecat.typename._
@@ -37,7 +37,7 @@ abstract class Mapping[F[_]](implicit val M: Monad[F]) extends QueryExecutor[F, 
   def compileAndRun(text: String, name: Option[String] = None, untypedVars: Option[Json] = None, introspectionLevel: IntrospectionLevel = Full, env: Env = Env.empty): F[Json] =
     compiler.compile(text, name, untypedVars, introspectionLevel) match {
       case Ior.Right(operation) =>
-        run(operation.query, schema.queryType, env)
+        run(operation.query, operation.rootTpe, env)
       case invalid =>
         QueryInterpreter.mkInvalidResponse(invalid).pure[F]
     }
@@ -58,10 +58,10 @@ abstract class Mapping[F[_]](implicit val M: Monad[F]) extends QueryExecutor[F, 
         }
     }
 
-  def rootCursor(path: List[String], rootTpe: Type, fieldName: String, child: Query, env: Env): F[Result[Cursor]] =
+  def rootCursor(path: List[String], rootTpe: Type, fieldName: String, child: Query, env: Env): F[Result[(Query, Cursor)]] =
     rootMapping(path, rootTpe, fieldName) match {
       case Some(root) =>
-        root.cursor(child, env)
+        root.run(child, env)
       case None =>
         mkErrorResult(s"No root field '$fieldName' in $rootTpe").pure[F]
     }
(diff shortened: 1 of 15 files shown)


modules/core/src/main/scala/mapping.scala
```diff
@@ -37,7 +37,7 @@ abstract class Mapping[F[_]](implicit val M: Monad[F]) extends QueryExecutor[F,
   def compileAndRun(text: String, name: Option[String] = None, untypedVars: Option[Json] = None, introspectionLevel: IntrospectionLevel = Full, env: Env = Env.empty): F[Json] =
     compiler.compile(text, name, untypedVars, introspectionLevel) match {
       case Ior.Right(operation) =>
-        run(operation.query, schema.queryType, env)
+        run(operation.query, operation.rootTpe, env)
```
tpolecat, line 40: This was a bug.

modules/skunk/src/test/scala/mutation/MutationData.scala
```diff
@@ -0,0 +1,141 @@
...
+              case Some((id, pop)) =>
+                pool.use { s =>
+                  s.prepare(sql"update city set population=$int4 where id=$int4".command).use { ps =>
+                    IorT.right(ps.execute(pop ~ id).void).value // awkward
+                  }
+                }
+              }
+          }),
+        ),
+      ),
```
tpolecat, line 94: The combined ergonomics of `Env` and `Result` and `F` make this a little awkward.

tpolecat:
Hm also some test failures. I'll look at those tomorrow.

tpolecat:
Ok it's working now. @milessabin I'm ready for your eyes when you have a moment. I'm not super happy with the ergonomics but that seems easy to fix with more `Mutation` constructors for common cases.

modules/core/src/main/scala/mapping.scala
milessabin: Good spot.

modules/skunk/src/test/scala/mutation/MutationData.scala
milessabin: Agreed, but this seems like a good starting point at least.

modules/skunk/src/test/scala/mutation/MutationData.scala
```diff
@@ -0,0 +1,141 @@
...
+            // We could also do this in the SqlRoot's mutation, and in fact would need to do so if
+            // the mutation generated a new id. But for now it seems easiest to do it here.
+            Unique(Eql(AttrPath(List("id")), Const(id)), child)
+          )
+        ).rightIor
+    }
+  ))
+}
+
+object MutationMapping extends SkunkMappingCompanion {
```
milessabin, line 134: Will this carry over to the doobie backend as well?

milessabin (approved):
LGTM, although it'd be nice to have doobie support as well.

tpolecat:
Cool, doobie support is on my checklist above, as well as a few more tests. Will ping you again for a final looksee when those are done.

modules/sql/src/test/scala/SqlMutationSpec.scala
```diff
@@ -0,0 +1,109 @@
...
+import org.scalatest.funsuite.AnyFunSuite
+import edu.gemini.grackle.QueryExecutor
+import cats.effect.IO
+import io.circe.Json
+import io.circe.literal._
+import edu.gemini.grackle.Schema
+
+trait SqlMutationSchema {
+
+  val schema =
```
tpolecat, line 15: It occurred to me that the schema really should be with the [shared] tests. Maybe I should change the rest of them.

tpolecat:
Alright I added an insert case that elaborates the `Select` at runtime and moved the test suite so it's shared with both database back-ends. Any last words?

milessabin:
This LGTM! :shipit:

PR #132, opened 2021-04-05
simple stream example

tpolecat:
The mechanism by which a `RootMapping`'s cursor is computed in `F` means that if `F` is `Stream[G, *]` then we can implement subscriptions trivially. There is probably more to this, by maybe not. In any case, for your consideration:


tpolecat:
Doesn't generalize. Subsumed by #133

PR #133, opened 2021-04-07
initial subscriptions implementation

tpolecat:
The strategy in #132 doesn't work in general ... we really do need to recognize that a GraphQL query can return many results in order to support subscriptions. Despite the changes here the existing API is mostly unbroken and the tests mostly compile and run unchanged.

- [x] change `QueryInterpreter` and underlying machinery (primarily `RootMapping`) to return streams.
- [x] figure out a way to make stats-collection work in the streaming case
- [ ] ~~add a test for a simple subscription driven by a queue~~ deferring this to a followup
- [x] add a test for a database subscription driven by a Postgres channel 

This also adds `SqlStatsMonitor` that collects stats in a `Ref`, allowing you to examine them after a stream has completed. It's minimal but sufficient for testing that the query compiler is doing the right thing, which is the immediate use case.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -5,6 +5,7 @@ val catsTestkitScalaTestVersion = "2.1.2"
 val circeVersion                = "0.13.0"
 val circeOpticsVersion          = "0.13.0"
 val doobieVersion               = "0.10.0"
+val fs2Version                  = "2.5.4"
 val http4sVersion               = "0.21.20"
 val kindProjectorVersion        = "0.11.3"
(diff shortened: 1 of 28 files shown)


tpolecat:
(rebased)

tpolecat:
The `subscription` operation doesn't quite work … I had to put the channel example on `query` to get it to work. So I need to track that down. But mechanically this seems to be working.

milessabin (approved):
LGTM!

PR #138, opened 2021-04-14
effectful value roots

tpolecat:
This changes `ValueRoot` roots from thunks to effectful values. The existing API is unchanged modulo strictness, which hasn't caused any issues in tests or the demo module. I also added some more constructors to the `Result` object.

The real point of this change was to allow me to write the included `SubscriptionSpec`, which doesn't rely on a database and thus can live in core.


milessabin (approved):
LGTM!

PR #140, opened 2021-04-14
schema and json literals

tpolecat:
This adds schema and json literals, the former for convenience and the latter to insulate us from a potentially drawn-out effort to get Circe literals into Scala 3.

Error reporting for schemas is bad because the parser errors are bad, and also because failure is Json blobs all the way down so there's no general way to pull out text messages, although knowing how it's implemented we could probably hack something up. Suggestions wanted here.

I only changed one of the tests to use the new literals, just to show that it works. Once we agree this is all ok I can change the rest and remove the circe-literal dependency.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -8,6 +8,7 @@ val doobieVersion               = "0.10.0"
 val fs2Version                  = "2.5.4"
 val http4sVersion               = "0.21.20"
 val kindProjectorVersion        = "0.11.3"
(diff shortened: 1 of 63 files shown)


tpolecat:
Ok with the changes in #141 the compile-time messages for invalid schemas look a lot better.

image

Without objection I'll update the rest of the schema literals and we can call this good.

milessabin (approved):
LGTM!

PR #141, opened 2021-04-15
add structured Problem type

tpolecat:
This changes `Result[+A]` to be `IorNec[Problem, A]` where `Problem` is a new data type containing message, locations, and path. `Problem` encodes to the same JSON structure we're generating now.

The use case for this is better error reporting for compile-time schema validation (#140).


milessabin (approved):
Yeah, that's much better :100:

PR #142, opened 2021-04-26
Unified fields and attributes; simplified DB nullability

milessabin:
+ Attributes are now fields with a `hidden` attribute.
+ There is no longer an attribute specific notion of path and all the logic which handled attribute vs. field paths has been eliminated.
+ Moved all DB nullability info onto the per-column `Codecs` which are now the sole source of field-mapping level nullability via `ColumnRef`.
+ Moved `TableDef` into the API, as part of `DoobieMapping`/`SkunkMapping`. Tests updated to use this style.


modules/core/src/main/scala/predicate.scala
```diff
@@ -14,16 +14,55 @@ import cats.implicits._
...
+ * we need to be able to construct where clauses from predicates over fields/attributes), so
+ * these cannot be arbitrary functions `Cursor => Boolean`.
+ */
+trait Term[T] extends Product with Serializable {
+  def apply(c: Cursor): Result[T]
+}
+
+sealed trait Path {
+  def path: List[String]
+  def extend(prefix: List[String]): Path
```
tpolecat, line 30: Reading the method name I expected it to take a suffix. Would something like `prefix` be a better name for this?

modules/core/src/main/scala/predicate.scala
```diff
@@ -14,16 +14,55 @@ import cats.implicits._
...
+  }
+
+  case class ListPath[T](val path: List[String]) extends Term[List[T]] with Path {
+    def apply(c: Cursor): Result[List[T]] =
+      c.flatListPath(path).map(_.map { case ScalarFocus(f: T @unchecked) => f })
+
+    def extend(prefix: List[String]): Path =
+      ListPath(prefix ++ path)
+  }
+}
```
tpolecat, line 54: Can you add a comment explaining when you would use `UniquePath` vs. `ListPath`? I see that both are used but I don't quite get the difference.

modules/core/src/main/scala/predicate.scala
milessabin: It's really "extend towards the root". Could you live with `prepend`?

modules/core/src/main/scala/predicate.scala
milessabin: :+1:

FWIW, none of this is new, just moved around slightly. Which doesn't affect your comment in any way :smile:

modules/core/src/main/scala/predicate.scala
tpolecat: Yes, I think `prepend` would be less confusing. To me anyway :-)

tpolecat:
> Moved `TableDef` into the API, as part of `DoobieMapping`/`SkunkMapping`.

I see this in `DoobieMapping` but not in `SkunkMapping`. Lost commit maybe?

tpolecat (commented):
This looks great. I love the simplification. A few comments but otherwise :+1:

tpolecat (approved):
LGTM, thanks!

PR #143, opened 2021-05-04
Scala 3

tpolecat:
resolves https://github.com/gemini-hlsw/gsp-graphql/issues/104

This adds a cross-build for Scala 3.0.0.

- Removed sbt-gsp (which doesn't work with Scala 3) and inlined the necessary settings.
- Added version-dependent Circe dependencies, with 0.13 for Scala 2 and 0.14 for Scala 3. They are not compatible and forcing Scala 2 people to use Circe 0.14 would likely break things. We can revisit this when things settle down a bit.
- Removed the dependency on circe-optics, which doesn't exist for Scala 3.  Workaround is `JsonExtractors` and a bit of `ACursor` bumbling in the tests.
- The inability to have a term of type `Codec[_]` forced me to write `ExistentialCodec` which pushes the type parameter down into a type member. The Anti-Aux pattern.
- `generic` is ported as-is, with all sources under `scala-2` an publishing disabled for Scala 3. We can add Scala 3 support in a followup PR.

TODO:
- [x] enable `demo` (http4s is available now)


modules/core/src/main/scala/compiler.scala
```diff
@@ -59,26 +59,23 @@ object QueryParser {
     }
   }
 
-  def parseOperation(op: Operation, fragments: Map[String, FragmentDefinition]): Result[UntypedOperation] = op match {
-    case Operation(opType, _, vds, _, sels) =>
-      val q = parseSelections(sels, None, fragments)
-      val vs = vds.map {
-        case VariableDefinition(nme, tpe, _) => UntypedVarDef(nme.value, tpe, None)
+  def parseOperation(op: Operation, fragments: Map[String, FragmentDefinition]): Result[UntypedOperation] = {
```
tpolecat, line 62: Scala 3 warns that this single-case match would only fail with `case null =>` so I collapsed it to a destructuring assignment. There are a few more of these.

modules/core/src/main/scala/jsonextractors.scala
```diff
@@ -0,0 +1,39 @@
+// Copyright (c) 2016-2021 Association of Universities for Research in Astronomy, Inc. (AURA)
+// For license information see LICENSE or https://opensource.org/licenses/BSD-3-Clause
+
+package edu.gemini.grackle
+
+import io.circe.Json
+import io.circe.JsonObject
+
+object JsonExtractor {
```
tpolecat, line 9: Because circe-optics doesn't exist for Scala 3 [yet].

modules/core/src/main/scala/predicate.scala
```diff
@@ -21,7 +21,7 @@ import QueryInterpreter.mkErrorResult
  * we need to be able to construct where clauses from predicates over fields/attributes), so
  * these cannot be arbitrary functions `Cursor => Boolean`.
  */
-trait Term[T] extends Product with Serializable {
+trait Term[T] extends Product with Serializable { // fun fact: making this covariant crashes Scala 3
```
tpolecat, line 24: Because `Term[+T]` crashes Scala there are some upcasts to `Term[Any]` later on.

modules/core/src/main/scala/predicate.scala
```diff
@@ -61,7 +61,10 @@ object Path {
     */
   case class ListPath[T](val path: List[String]) extends Term[List[T]] with Path {
     def apply(c: Cursor): Result[List[T]] =
-      c.flatListPath(path).map(_.map { case ScalarFocus(f: T @unchecked) => f })
+      c.flatListPath(path).map(_.map {
+        case ScalarFocus(f: T @unchecked) => f
+        case _ => sys.error("impossible")
```
tpolecat, line 66: The exhaustiveness checker is a lot better so I added a bunch of these. Need to go back and improve these.

modules/core/src/main/scala/queryinterpreter.scala
```diff
@@ -170,7 +170,7 @@ class QueryInterpreter[F[_]](mapping: Mapping[F]) {
             case Ior.Right(elem) => (errors, elem :: elems)
             case Ior.Both(errs, elem) => (errs.toChain ++ errors, elem :: elems)
           }
-      }).map(_.reverse)
+      }).fmap(_.reverse)
```
tpolecat, line 173: Functor `.map` syntax doesn't work for tuples in Scala 3 because there's polymorphic `map` on `Tuple` now.

modules/core/src/main/scala/schema.scala
```diff
@@ -630,7 +630,7 @@ case class NullableType(
  *
  * @see https://facebook.github.io/graphql/draft/#sec-The-__Field-Type
  */
-case class Field private(
+case class Field(
```
tpolecat, line 633: We're calling this from a place where Scala 3 says we can't. I don't think it needs to be `private` anyway.

modules/core/src/test/scala/compiler/Environment.scala
```diff
@@ -1,10 +1,11 @@
-// Copyright (c) 2016-2020 Association of Universities for Research in Astronomy, Inc. (AURA)
+// Copyright (c) 2016-2021 Association of Universities for Research in Astronomy, Inc. (AURA)
 // For license information see LICENSE or https://opensource.org/licenses/BSD-3-Clause
 
 package compiler
 
 import cats.Id
 import cats.implicits._
+import cats.catsInstancesForId
```
tpolecat, line 8: Scala 3 can't find instances for `cats.Id` but it's not clear why. This fixes it.

modules/doobie/src/main/scala/DoobieMapping.scala
```diff
@@ -12,14 +12,15 @@ import _root_.doobie.enumerated.Nullability.{Nullable, NoNulls}
...
+import scala.annotation.unchecked.uncheckedVariance
 
 abstract class DoobieMapping[F[_]: Sync](
   val transactor: Transactor[F],
   val monitor:    DoobieMonitor[F]
 ) extends SqlMapping[F] {
 
   type Codec[A]   = (Meta[A], Boolean)
-  type Encoder[A] = (Put[A], Boolean)
+  type Encoder[-A] = (Put[A], Boolean) @uncheckedVariance
```
tpolecat, line 23: Encoder has to be contravariant or `MappedQuery` has one million pages of type errors.

modules/sql/src/main/scala/SqlMapping.scala
```diff
@@ -190,7 +190,7 @@ trait SqlMapping[F[_]] extends CirceMapping[F] with SqlModule[F] { self =>
   final class MappedQuery(
     table: String,
     columns: List[ColumnRef],
-    metas: List[(Boolean, Codec[_])],
+    metas: List[(Boolean, ExistentialCodec)],
```
tpolecat, line 193: `ExistentialCodec` is defined in `SqlModule`, see below.

modules/sql/src/main/scala/SqlModule.scala
```diff
@@ -14,8 +14,24 @@ trait SqlModule[F[_]] {
   /** The type of a codec that reads and writes column values of type `A`. */
   type Codec[A]
 
+  /** A codec that has forgotten its type argument. */
```
tpolecat, line 17: `Codec[_]` isn't a type in Scala 3 so we have to push the existential down into a type member.

modules/sql/src/main/scala/SqlModule.scala
```diff
@@ -14,8 +14,24 @@ trait SqlModule[F[_]] {
...
+        val codec = c
+      }
+  }
+
+  implicit def unliftCodec(c: ExistentialCodec): Codec[c.A] =
+    c.codec
+
   /** The type of an encoder that writes column values of type `A`. */
-  type Encoder[A]
+  type Encoder[-A]
```
tpolecat, line 34: If `Encoder` is invariant there are one million pages of type errors in `MappedQuery`. Scala 2 is happy to crush stuff to `Any` after a while but Scala 3 keeps up and doesn't like what's happening in calls to `loop(...)`.

build.sbt
```diff
@@ -82,13 +96,6 @@ lazy val circe = project
   .settings(commonSettings)
   .settings(
     name := "gsp-graphql-circe",
-    libraryDependencies ++= Seq(
-      "org.tpolecat"      %% "atto-core"              % attoVersion,
-      "org.typelevel"     %% "cats-core"              % catsVersion,
-      "io.circe"          %% "circe-core"             % circeVersion,
-      "io.circe"          %% "circe-optics"           % circeOpticsVersion,
-      "io.circe"          %% "circe-parser"           % circeVersion
-    )
```
tpolecat, line 91: These dependencies all arrive transitively so I removed them as a simplification. Same in a few places below.

docs/src/main/paradox/tutorial/in-memory-model.md
```diff
@@ -193,7 +193,7 @@ Here the root of the result is "Luke Skywalker" and we loop back to Luke through
 The data type is not itself recursive, instead friend are identified by their ids. When traversing through the list of
 friends the `resolveFriends` method is used to locate the next `Character` value to visit.
```
tpolecat, line 194: No idea why this is being rendered in red.

modules/circe/src/test/scala/CirceData.scala
```diff
@@ -5,6 +5,7 @@ package edu.gemini.grackle
 package circetests
 
 import cats.Id
+import cats.catsInstancesForId
```
tpolecat, line 8: These instances are unavailable in Scala 3. I don't know why.

modules/core/src/main/scala/parser.scala
```diff
@@ -189,7 +189,7 @@ object GraphQLParser {
     QueryShorthand.widen[Ast.OperationDefinition] | Operation.widen
 
   lazy val QueryShorthand: Parser[Ast.OperationDefinition.QueryShorthand] =
-    SelectionSet.map(Ast.OperationDefinition.QueryShorthand)
+    SelectionSet.map(Ast.OperationDefinition.QueryShorthand.apply)
```
tpolecat, line 192: Scala 3 warns that `apply`-insertion is deprecated. I guess the companions no longer extend function types?

milessabin (approved):
LGTM!

tpolecat:
Waiting on test-containers, which is having some issues. https://github.com/testcontainers/testcontainers-scala/pull/177

tpolecat:
test-containers is fixed, now waiting on cats-testkit-scalatest

tpolecat:
(updated to 3.0.0 final)

PR #147, opened 2021-05-27
Use shapeless 3 for generic module for Scala 3

milessabin:


modules/generic/src/main/scala/genericmapping.scala
```diff
@@ -0,0 +1,231 @@
...
+      override def asLeaf: Result[Json] = Json.fromBoolean(focus).rightIor
+    }
+    new CursorBuilder[Boolean] {
+      val tpe = BooleanType
+      def build(path: List[String], focus: Boolean, parent: Option[Cursor], env: Env): Result[Cursor] =
+        BooleanCursor(path, focus, parent, env).rightIor
+    }
+  }
+
+  implicit def deriveEnumerationCursorBuilder[T <: Enumeration#Value](tpe0: Type): CursorBuilder[T] = {
```
tpolecat, line 133: Should this be `implicit`?

tpolecat (approved):
Looks great, thanks!

modules/generic/src/main/scala/genericmapping.scala
milessabin: No it shouldn't ... good spot :smile:

PR #150, opened 2021-06-04
Fixed prefixed mapping filter issue

milessabin:
+ A bug in the lookup for an encoder corresponding to a field under a prefixed mapping could result in the encoder not being found and filters referring to that term not being translated to SQL where clauses.
+ Added a test for rows fetched for a filtered (rather than unique) query.
+ Also fixed a bug in `TableDef#col` where a bogus `TypeName` instance could be summoned out of thin air.


TimWSpence (approved):
Thanks for fixing so quickly! :)

PR #152, opened 2021-07-10
Rework SQL compiler

milessabin:
+ Complete rewrite of the SQL compiler.
+ Removed SQL staging logic now that GraphQL queries are mapped to single SQL queries. It's conceivable we might and to reintroduce this later to support non-Postgres DBs.
+ Factored out compiler/interpreter position into a `Context` type which holds path, result path and type.
+ `Unique` now just converts `[T]` to `T`.
+ Removed `Project` now that multi-step predicate paths are now supported correctly.
+ Removed `GroupBy` query algebra term which was only used for (now removed) staging.
+ `Codec`/`Encoder` type parameter removed.
+ Added order independent result testing utility.
+ Bumped dependencies to most recent CE2 versions.
+ Added local Docker compose for experimentation.


milessabin:
Rebased to pick up the commit missed from main :roll_eyes:

tpolecat (approved):
This looks great! Looking forward to trying it out :-)

PR #156, opened 2021-07-13
Fixed skunk handling of nullable columns in failed joins

tpolecat:
Ok so the deal is that `SkunkMapping` will interpret _any_ null in a joined column as `FailedJoin`, even if the column is nullable, because we don't have any type information at that point. So we can compute nullity and pass it in, or I can add a flag to `Codec` and friends in Skunk.


milessabin (approved):
LGTM!

PR #158, opened 2021-07-22
Rewrite parser using cats-parse

### LINKED ISSUE #157: support multi-line strings in parser
The parser currently doesn't support multi-line strings, like

```graphql
"""
The dreaded Woozle.
"""
type Woozle {
  ...
}
```

TimWSpence:
resolves #157

### DIFF
--- a/modules/core/src/main/scala/compiler.scala
+++ b/modules/core/src/main/scala/compiler.scala
@@ -5,8 +5,8 @@ package edu.gemini.grackle
 
 import scala.annotation.tailrec
 
-import atto.Atto._
 import cats.data.Ior
+import cats.parse.Parser
 import cats.implicits._
 import io.circe.Json
 
@@ -27,11 +27,11 @@ object QueryParser {
    *  Yields a Query value on the right and accumulates errors on the left.
    */
   def parseText(text: String, name: Option[String] = None): Result[UntypedOperation] = {
-    def toResult[T](pr: Either[String, T]): Result[T] =
-      Ior.fromEither(pr).leftMap(mkOneError(_))
+    def toResult[T](pr: Either[Parser.Error, T]): Result[T] =
+      Ior.fromEither(pr).leftMap(e => mkOneError(e.expected.toList.mkString(",")))
 
     for {
-      doc   <- toResult(GraphQLParser.Document.parseOnly(text).either)
+      doc   <- toResult(GraphQLParser.Document.parseAll(text))
       query <- parseDocument(doc, name)
     } yield query
   }
@@ -173,7 +173,7 @@ object QueryMinimizer {
 
   def minimizeText(text: String): Either[String, String] = {
     for {
-      doc <- GraphQLParser.Document.parseOnly(text).either
+      doc <- GraphQLParser.Document.parseAll(text).leftMap(_.expected.toList.mkString(","))
     } yield minimizeDocument(doc)
   }
 
--- a/modules/core/src/main/scala/parser.scala
+++ b/modules/core/src/main/scala/parser.scala
@@ -349,8 +349,8 @@ object GraphQLParser {
     keyword("@") *> (Name ~ opt(Arguments)).map { case (n, ods) => Ast.Directive(n, ods.orEmpty)}
 
   lazy val Name: Parser[Ast.Name] = {
-    val initial    = ('A' to 'Z').toSet ++ ('a' to 'z').toSet + '_'
-    val subsequent = initial ++ ('0' to '9').toSet
+    val initial    = ('A' to 'Z') ++ ('a' to 'z') ++ Seq('_')
+    val subsequent = initial ++ ('0' to '9')
     token(charIn(initial) ~ many(charIn(subsequent))).map {
       case (h, t) => Ast.Name((h :: t).mkString)
     }
@@ -369,9 +369,8 @@ object CommentedText {
     charsWhile0(c => c.isWhitespace || c == ',').void.withContext("whitespace")
 
   /** Parser that consumes a comment */
-  def comment: Parser[Unit] = {
+  def comment: Parser[Unit] =
     (char('#') *> many((charWhere(c => c != '\n' && c != '\r'))) <* charIn('\n', '\r') <* skipWhitespace).void.withContext("comment")
-  }
 
   /** Turns a parser into one that skips trailing whitespace and comments */
   def token[A](p: Parser[A]): Parser[A] =
--- a/modules/core/src/main/scala/parser.scala
+++ b/modules/core/src/main/scala/parser.scala
@@ -22,7 +22,7 @@ object GraphQLParser {
     ExecutableDefinition | TypeSystemDefinition // | TypeSystemExtension
 
   lazy val TypeSystemDefinition: Parser[Ast.TypeSystemDefinition] =
-    SchemaDefinition | TypeDefinition | DirectiveDefinition
+    SchemaDefinition | TypeDefinition.backtrack | DirectiveDefinition
 
   lazy val SchemaDefinition: Parser[Ast.SchemaDefinition] =
     for {
@@ -39,11 +39,11 @@ object GraphQLParser {
     } yield Ast.RootOperationTypeDefinition(optpe, tpe)
 
   lazy val TypeDefinition: Parser[Ast.TypeDefinition] =
-    ScalarTypeDefinition.widen[Ast.TypeDefinition] |
-    ObjectTypeDefinition |
-    InterfaceTypeDefinition |
-    UnionTypeDefinition |
-    EnumTypeDefinition |
+    ScalarTypeDefinition.widen[Ast.TypeDefinition].backtrack |
+    ObjectTypeDefinition.backtrack |
+    InterfaceTypeDefinition.backtrack |
+    UnionTypeDefinition.backtrack |
+    EnumTypeDefinition.backtrack |
     InputObjectTypeDefinition
 
   lazy val ScalarTypeDefinition: Parser[Ast.ScalarTypeDefinition] =
--- a/modules/core/src/main/scala/parser.scala
+++ b/modules/core/src/main/scala/parser.scala
@@ -433,8 +433,8 @@ object Literals {
 
     lazy val bigDecimal: Parser[BigDecimal] = for {
       a <- intLiteral
-      b <- (char('.') *> digit.rep).string.?
-      c <- ((char('e') | char('E')) *> intLiteral).string.?
+      b <- (char('.') *> digit.rep.string).?
+      c <- ((char('e') | char('E')) *> intLiteral.string).?
       res <- (a,b,c) match {
         case (a, Some(b), None) => pure(BigDecimal(s"$a.$b"))
         case (a, None, Some(c)) => pure(BigDecimal(s"${a}E$c"))
--- a/modules/core/src/main/scala/parser.scala
+++ b/modules/core/src/main/scala/parser.scala
@@ -224,16 +224,16 @@ object GraphQLParser {
 
     val NumericLiteral: Parser[Ast.Value] = {
 
-      def narrow(d: BigDecimal): Parser0[Ast.Value.FloatValue] =
-        pure(Ast.Value.FloatValue(d.toDouble))
+      def narrow(d: BigDecimal): Ast.Value.FloatValue =
+        Ast.Value.FloatValue(d.toDouble)
 
       token(
         (intLiteral ~ (char('.') *> digit.rep.string).? ~ ((char('e') | char('E')) *> intLiteral.string).?)
-          .flatMap {
+          .map {
             case ((a, Some(b)), None) => narrow(BigDecimal(s"$a.$b"))
             case ((a, None), Some(c)) => narrow(BigDecimal(s"${a}E$c"))
             case ((a, Some(b)), Some(c)) => narrow(BigDecimal(s"$a.${b}E$c"))
-            case ((a, None), None) => pure(Ast.Value.IntValue(a))
+            case ((a, None), None) => Ast.Value.IntValue(a)
           }
       )
     }
--- a/modules/core/src/main/scala/parser.scala
+++ b/modules/core/src/main/scala/parser.scala
@@ -259,7 +259,38 @@ object GraphQLParser {
   lazy val TypeCondition: Parser[Ast.Type.Named] =
     keyword("on") *> NamedType
 
-  lazy val Value: Parser[Ast.Value] =
+  lazy val Value: Parser[Ast.Value] = recursive[Ast.Value] { rec =>
+
+    lazy val NullValue: Parser[Ast.Value.NullValue.type] =
+      keyword("null").as(Ast.Value.NullValue)
+
+    lazy val EnumValue: Parser[Ast.Value.EnumValue] =
+      Name.filter {
+        case Ast.Name("true")  => false
+        case Ast.Name("false") => false
+        case Ast.Name("null")  => false
+        case _                 => true
+      } .map(Ast.Value.EnumValue.apply)
+
+    //TODO this is recursive and SOs on construction of the parser
+    lazy val ListValue: Parser[Ast.Value.ListValue] =
+      token(squareBrackets(many(rec)).map(Ast.Value.ListValue.apply))
+
+    lazy val IntValue: Parser[Ast.Value.IntValue] =
+      token(intLiteral).map(a => Ast.Value.IntValue(a.toInt))
+
+    lazy val FloatValue: Parser[Ast.Value.FloatValue] =
+      token(floatLiteral).map(a => Ast.Value.FloatValue(a.toDouble))
+
+    lazy val BooleanValue: Parser[Ast.Value.BooleanValue] =
+      token(booleanLiteral).map(Ast.Value.BooleanValue.apply)
+
+    lazy val ObjectValue: Parser[Ast.Value.ObjectValue] =
+      braces(many(ObjectField)).map(Ast.Value.ObjectValue.apply)
+
+    lazy val ObjectField: Parser[(Ast.Name, Ast.Value)] =
+      (Name <* keyword(":")) ~ rec
+
     Variable    .widen[Ast.Value] |
     IntValue    .widen            |
     FloatValue  .widen            |
@@ -269,38 +300,10 @@ object GraphQLParser {
     EnumValue   .widen            |
     ListValue   .widen            |
     ObjectValue .widen
-
-  lazy val NullValue: Parser[Ast.Value.NullValue.type] =
-    keyword("null").as(Ast.Value.NullValue)
-
-  lazy val EnumValue: Parser[Ast.Value.EnumValue] =
-    Name.filter {
-      case Ast.Name("true")  => false
-      case Ast.Name("false") => false
-      case Ast.Name("null")  => false
-      case _                 => true
-    } .map(Ast.Value.EnumValue.apply)
-
-  lazy val ListValue: Parser[Ast.Value.ListValue] =
-    token(squareBrackets(many(Value)).map(Ast.Value.ListValue.apply))
-
-  lazy val IntValue: Parser[Ast.Value.IntValue] =
-    token(intLiteral).map(a => Ast.Value.IntValue(a.toInt))
-
-  lazy val FloatValue: Parser[Ast.Value.FloatValue] =
-    token(floatLiteral).map(a => Ast.Value.FloatValue(a.toDouble))
+  }
 
   lazy val StringValue: Parser[Ast.Value.StringValue] =
     token(stringLiteral).map(Ast.Value.StringValue.apply)
-
-  lazy val BooleanValue: Parser[Ast.Value.BooleanValue] =
-    token(booleanLiteral).map(Ast.Value.BooleanValue.apply)
-
-  lazy val ObjectValue: Parser[Ast.Value.ObjectValue] =
-    braces(many(ObjectField)).map(Ast.Value.ObjectValue.apply)
-
-  lazy val ObjectField: Parser[(Ast.Name, Ast.Value)] =
-    (Name <* keyword(":")) ~ Value
 
   lazy val VariableDefinitions: Parser[List[Ast.VariableDefinition]] =
     parens(many(VariableDefinition))
@@ -335,8 +338,9 @@ object GraphQLParser {
   lazy val Directives: Parser0[List[Ast.Directive]] =
     many(Directive)
 
+  //TODO why does this hang?
   lazy val Directive: Parser[Ast.Directive] =
-    keyword("@") *> (Name, opt(Arguments)).mapN((n, ods) => Ast.Directive(n, ods.orEmpty))
+    keyword("@") *> (Name ~ opt(Arguments)).map { case (n, ods) => Ast.Directive(n, ods.orEmpty)}
 
   lazy val Name: Parser[Ast.Name] = {
     val initial    = ('A' to 'Z').toSet ++ ('a' to 'z').toSet + '_'
@@ -355,12 +359,12 @@ object CommentedText {
 
   def many[A](parser: Parser[A]): Parser0[List[A]] = parser.rep0
 
-  def skipWhitespace: Parser[Unit] =
-    charsWhile(c => c.isWhitespace || c == ',').void.withContext("whitespace")
+  def skipWhitespace: Parser0[Unit] =
+    charsWhile0(c => c.isWhitespace || c == ',').void.withContext("whitespace")
 
   /** Parser that consumes a comment */
   def comment: Parser[Unit] = {
-    (char('#') *> many((charWhere(c => c != '\n' && c != '\r'))) <* charIn('\n', '\r') <* skipWhitespace).void
+    (char('#') *> many((charWhere(c => c != '\n' && c != '\r'))) <* charIn('\n', '\r') <* skipWhitespace).void.withContext("comment")
   }
 
   /** Turns a parser into one that skips trailing whitespace and comments */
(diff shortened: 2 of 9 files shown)


TimWSpence:
Opening this to get feedback on the parser changes. The backtracking in particular feels quite arbitrary and unsatisfying at the moment - would love feedback on a more idiomatic way to structure it.

milessabin:
@johnynek we'd be very grateful for your input on this ...

modules/core/src/main/scala/compiler.scala
```diff
@@ -173,7 +173,7 @@ object QueryMinimizer {
 
   def minimizeText(text: String): Either[String, String] = {
     for {
-      doc <- GraphQLParser.Document.parseOnly(text).either
+      doc <- GraphQLParser.Document.parseAll(text).leftMap(_.expected.toList.mkString(","))
```
johnynek, line 176: while this may be okay for the short term, I would probably make a custom `Parser.Error => String` function can call that (presumably also with access to `text` and possibly using `cats.parse.LocationMap`).

modules/core/src/main/scala/parser.scala
```diff
@@ -312,71 +316,145 @@ object GraphQLParser {
...
   }
 }
 
 object CommentedText {
 
-  def skipWhitespace: Parser[Unit] =
-    takeWhile(c => c.isWhitespace || c == ',').void named "whitespace"
+  lazy val whitespace: Parser[Char] = charWhere(_.isWhitespace)
+
+  def opt[A](parser: Parser0[A]): Parser0[Option[A]] = parser.backtrack.?
```
johnynek, line 364: hiding this backtrack in here I wouldn't do. Honestly, I would just use `.?` in the api rather than using `opt` here.

The problem with backtrack, as you may know, is that it risks hiding errors and blowing up parsing times. Most parsers should be written so that they consume nothing or pass, partial failures should be only for some tricky narrow spots in the grammar.

modules/core/src/main/scala/parser.scala
```diff
@@ -312,71 +316,145 @@ object GraphQLParser {
...
+
+  def many[A](parser: Parser[A]): Parser0[List[A]] = parser.rep0
+
+  def skipWhitespace: Parser0[Unit] =
+    charsWhile0(c => c.isWhitespace || c == ',').void.withContext("whitespace")
 
   /** Parser that consumes a comment */
   def comment: Parser[Unit] = {
```
johnynek, line 372: nit: you don't need `{` here.

modules/core/src/main/scala/parser.scala
```diff
@@ -312,71 +316,145 @@ object GraphQLParser {
...
+  def token0[A](p: Parser0[A]): Parser0[A] =
+    p <* skipWhitespace <* many(comment)
 
   /**
    * Consumes `left` and `right`, including the trailing and preceding whitespace,
    * respectively, and returns the value of `p`.
    */
-  private def _bracket[A,B,C](left: Parser[B], p: => Parser[A], right: => Parser[C]): Parser[A] =
-    token(left) ~> token(p) <~ token(right)
+  private def _bracket[A,B,C](left: Parser[B], p: => Parser0[A], right: => Parser[C]): Parser[A] =
```
johnynek, line 387: I don't see any reason for the parameters here to be call by name.

modules/core/src/main/scala/parser.scala
```diff
@@ -312,71 +316,145 @@ object GraphQLParser {
...
+    token(left) *> token0(p) <* token(right)
 
   /** Turns a parser into one that consumes surrounding parentheses `()` */
-  def parens[A](p: => Parser[A]): Parser[A] =
-    _bracket(char('('), p, char(')')).named(s"parens(${p.toString})")
+  def parens[A](p: => Parser0[A]): Parser[A] =
```
johnynek, line 391: don't need call by name.

modules/core/src/main/scala/parser.scala
```diff
@@ -312,71 +316,145 @@ object GraphQLParser {
...
+    _bracket(char('('), p, char(')')).withContext(s"parens(${p.toString})")
```
johnynek, line 392: question: do you not allow whitespace after the parens here?

modules/core/src/main/scala/parser.scala
```diff
@@ -312,71 +316,145 @@ object GraphQLParser {
...
 
   /** Turns a parser into one that consumes surrounding curly braces `{}` */
-  def braces[A](p: => Parser[A]): Parser[A] = {
+  def braces[A](p: => Parser0[A]): Parser[A] = {
```
johnynek, line 395: don't need call by name here. I think it is being forced.

If you need laziness, using `Parser.defer` might be more explicit.

modules/core/src/main/scala/parser.scala
```diff
@@ -312,71 +316,145 @@ object GraphQLParser {
...
     lazy val q = p
-    _bracket(char('{'), q, char('}')).named(s"braces(${q.toString})")
+    _bracket(char('{'), q, char('}')).withContext(s"braces(${q.toString})")
   }
 
   /** Turns a parser into one that consumes surrounding square brackets `[]` */
-  def squareBrackets[A](p: => Parser[A]): Parser[A] = {
+  def squareBrackets[A](p: => Parser0[A]): Parser[A] = {
```
johnynek, line 401: no need call by name

modules/core/src/main/scala/parser.scala
```diff
@@ -312,71 +316,145 @@ object GraphQLParser {
...
     lazy val q = p
-    _bracket(char('['), q, char(']')).named(s"squareBrackets(${q.toString})")
+    _bracket(char('['), q, char(']')).withContext(s"squareBrackets(${q.toString})")
+  }
+}
+
+object Literals {
+
+  lazy val stringLiteral: Parser[String] = {
```
johnynek, line 409: I think you should not default to `lazy val` and should instead only use it when you really need it. Looks like everything is getting lazy val here.

modules/core/src/main/scala/parser.scala
```diff
@@ -312,71 +316,145 @@ object GraphQLParser {
...
+      (peek(not(charIn('"', '\\') | lineTerminator)).with1 *> sourceCharacter) |
+        (string("\\u") ~ escapedUnicode).string |
+        (char('\\') ~ escapedCharacter).string
+    )
+
+    lazy val blockStringCharacter = string("TODO").string
+
+    //TODO should this be a Parser[Char] which converts to unicode? I guess so?
+    //See Atto for reference
+    lazy val escapedUnicode: Parser[String] = (digit | hexdig).repExactlyAs[String](4)
```
johnynek, line 421: you can go both ways. One approach you might want is to parse a faithful representation of valid input, then convert at a high layer. That may not be needed here.

If not, you can just write `Parser[Char]` that parses escape codes and then maps them into the unicode character, then use `repAs` or to get your final resulting string.

modules/core/src/main/scala/parser.scala
```diff
@@ -312,71 +316,145 @@ object GraphQLParser {
...
+
+    lazy val escapedCharacter = charIn('"', '\\', '/', 'b', 'f', 'n', 'r', 't')
+
+    lazy val sourceCharacter: Parser[String] = (charIn(0x0009.toChar, 0x000A.toChar, 0x000D.toChar) | charIn(0x0020.toChar to 0xFFFF.toChar)).string
+
+    lazy val lineTerminator = (lf | cr | crlf).string
+
+    stringCharacter.rep0.string.with1.surroundedBy(char('"')) | (blockStringCharacter.rep0.string.with1.surroundedBy(string("\"\"\"")))
+
```
johnynek, line 430: you can compare this approach with some code I wrote here:

https://github.com/typelevel/cats-parse/blob/85f3c7f3ee67f9fe96c7754bd06e81a00d533da1/bench/src/main/scala/cats/parse/bench/self.scala#L130

this code was designed for a python like language that used either `'` or `"` as the start of your strings and can be simplified somewhat if you have a single approach and set of escapes.

modules/core/src/main/scala/parser.scala
```diff
@@ -312,71 +316,145 @@ object GraphQLParser {
...
+  }
+
+  lazy val intLiteral: Parser[Int] =
+    bigInt.flatMap {
+      case v if v.isValidInt => pure(v.toInt)
+      case v => failWith(s"$v is larger than max int")
+    }
+
+  lazy val booleanLiteral: Parser[Boolean] = string("true").as(true) | string("false").as(false)
```
johnynek, line 439: no need for lazy

modules/core/src/main/scala/parser.scala
```diff
@@ -312,71 +316,145 @@ object GraphQLParser {
...
+
+    lazy val sourceCharacter: Parser[String] = (charIn(0x0009.toChar, 0x000A.toChar, 0x000D.toChar) | charIn(0x0020.toChar to 0xFFFF.toChar)).string
+
+    lazy val lineTerminator = (lf | cr | crlf).string
+
+    stringCharacter.rep0.string.with1.surroundedBy(char('"')) | (blockStringCharacter.rep0.string.with1.surroundedBy(string("\"\"\"")))
+
+  }
+
+  lazy val intLiteral: Parser[Int] =
```
johnynek, line 433: no need for lazy

modules/core/src/main/scala/parser.scala
```diff
@@ -312,71 +316,145 @@ object GraphQLParser {
...
+    bigInt.flatMap {
+      case v if v.isValidInt => pure(v.toInt)
+      case v => failWith(s"$v is larger than max int")
+    }
+
+  lazy val booleanLiteral: Parser[Boolean] = string("true").as(true) | string("false").as(false)
+
+  lazy val floatLiteral: Parser[Float] = {
+
+    lazy val bigDecimal: Parser[BigDecimal] = for {
```
johnynek, line 443: this use of flatMap here is going to be considerably slower for float heavy code, I think.

You could write your parser with only 1 flatMap easily:
```scala
(intLiteral, (char('.') *> digit.rep.string).?, ((char('e') | char('E')) *> intLiteral.string).?)
  .tupled
  .flatMap {
        case (a, Some(b), None) => pure(BigDecimal(s"$a.$b"))
        case (a, None, Some(c)) => pure(BigDecimal(s"${a}E$c"))
        case (a, Some(b), Some(c)) => pure(BigDecimal(s"$a.${b}E$c"))
        case (a, None, None) => failWith(s"$a is not a valid float - must have at least one of a fractional or exponent part") 
  }
```
that said, I think if you compare to this:

https://github.com/typelevel/cats-parse/blob/85f3c7f3ee67f9fe96c7754bd06e81a00d533da1/core/shared/src/main/scala/cats/parse/Numbers.scala#L65

you can see a way to write this with no flatMap.

modules/core/src/main/scala/parser.scala
```diff
@@ -312,71 +316,145 @@ object GraphQLParser {
...
+
+  lazy val Directives: Parser0[List[Ast.Directive]] =
     many(Directive)
 
   lazy val Directive: Parser[Ast.Directive] =
-    keyword("@") ~> (Name, opt(Arguments)).mapN((n, ods) => Ast.Directive(n, ods.orEmpty))
+    keyword("@") *> (Name ~ opt(Arguments)).map { case (n, ods) => Ast.Directive(n, ods.orEmpty)}
 
   lazy val Name: Parser[Ast.Name] = {
     val initial    = ('A' to 'Z').toSet ++ ('a' to 'z').toSet + '_'
```
johnynek, line 352: nit: you don't need the `.toSet` on the `('a' to 'z')` you just need an iterable

modules/core/src/main/scala/parser.scala
```diff
@@ -251,13 +253,43 @@ object GraphQLParser {
...
+  lazy val TypeCondition: Parser[Ast.Type.Named] =
+    keyword("on") *> NamedType
+
+  lazy val Value: Parser[Ast.Value] = recursive[Ast.Value] { rec =>
+
+    lazy val NullValue: Parser[Ast.Value.NullValue.type] =
+      keyword("null").as(Ast.Value.NullValue)
+
+    lazy val EnumValue: Parser[Ast.Value.EnumValue] =
+      Name.filter {
```
johnynek, line 268: I think it would be bettter to have `Name` not parse true, false or null. so you don't have a consuming failure here (which filter will create).

`not(string("true") | string("false") | string("null")) *> ...` could be how `Name` is defined.

modules/core/src/main/scala/parser.scala
```diff
@@ -102,53 +104,53 @@ object GraphQLParser {
...
+  lazy val ArgumentsDefinition: Parser[List[Ast.InputValueDefinition]] =
     parens(many(InputValueDefinition))
 
-  lazy val InputFieldsDefinition =
+  lazy val InputFieldsDefinition: Parser[List[Ast.InputValueDefinition]] =
     braces(many(InputValueDefinition))
 
-  lazy val InputValueDefinition =
+  lazy val InputValueDefinition: Parser[Ast.InputValueDefinition] =
     for {
```
johnynek, line 129: you are much better off not using flatMap but instaed `mapN` here.

modules/core/src/main/scala/parser.scala
```diff
@@ -102,53 +104,53 @@ object GraphQLParser {
...
-    (keyword("implements") ~ opt(keyword("&"))) ~> sepBy(NamedType, keyword("&"))
+    (keyword("implements") ~ opt(keyword("&"))) *> NamedType.repSep0(keyword("&"))
 
-  lazy val FieldsDefinition =
+  lazy val FieldsDefinition: Parser[List[Ast.FieldDefinition]] =
     braces(many(FieldDefinition))
 
-  lazy val FieldDefinition =
+  lazy val FieldDefinition: Parser[Ast.FieldDefinition] =
     for {
```
johnynek, line 113: don't use flatMap but `mapN` here.

modules/core/src/main/scala/parser.scala
```diff
@@ -3,96 +3,98 @@
...
+      desc   <- opt(Description).with1
       _      <- keyword("enum")
       name   <- Name
       dirs   <- opt(Directives)
       values <- EnumValuesDefinition
     } yield Ast.EnumTypeDefinition(name, desc.map(_.value), dirs.getOrElse(Nil), values)
 
-  lazy val InputObjectTypeDefinition =
+  lazy val InputObjectTypeDefinition: Parser[Ast.InputObjectTypeDefinition] =
     for {
```
johnynek, line 96: don't use flatMap but mapN here

modules/core/src/main/scala/parser.scala
```diff
@@ -3,96 +3,98 @@
...
+      desc   <- opt(Description).with1
       _      <- keyword("union")
       name   <- Name
       dirs   <- opt(Directives)
       members <- UnionMemberTypes
     } yield Ast.UnionTypeDefinition(name, desc.map(_.value), dirs.getOrElse(Nil), members)
 
-  lazy val EnumTypeDefinition =
+  lazy val EnumTypeDefinition: Parser[Ast.EnumTypeDefinition] =
     for {
```
johnynek, line 87: use mapN instead of flatMap.

modules/core/src/main/scala/parser.scala
```diff
@@ -3,96 +3,98 @@
...
       _      <- keyword("interface")
       name   <- Name
       ifs    <- opt(ImplementsInterfaces)
       dirs   <- opt(Directives)
       fields <- FieldsDefinition
     } yield Ast.InterfaceTypeDefinition(name, desc.map(_.value), fields, ifs.getOrElse(Nil), dirs.getOrElse(Nil))
 
-  lazy val UnionTypeDefinition =
+  lazy val UnionTypeDefinition: Parser[Ast.UnionTypeDefinition] =
     for {
```
johnynek, line 78: replace flatMap with mapN

modules/core/src/main/scala/parser.scala
```diff
@@ -3,96 +3,98 @@
...
       _      <- keyword("type")
       name   <- Name
       ifs    <- opt(ImplementsInterfaces)
       dirs   <- opt(Directives)
       fields <- FieldsDefinition
     } yield Ast.ObjectTypeDefinition(name, desc.map(_.value), fields, ifs.getOrElse(Nil), dirs.getOrElse(Nil))
 
-  lazy val InterfaceTypeDefinition =
+  lazy val InterfaceTypeDefinition: Parser[Ast.InterfaceTypeDefinition] =
     for {
```
johnynek, line 68: replace flatMap with mapN

modules/core/src/main/scala/parser.scala
```diff
@@ -3,96 +3,98 @@
...
-      desc   <- opt(Description)
+      desc   <- opt(Description).with1
       _      <- keyword("scalar")
       name   <- Name
       dirs   <- opt(Directives)
     } yield Ast.ScalarTypeDefinition(name, desc.map(_.value), dirs.getOrElse(Nil))
 
-  lazy val ObjectTypeDefinition =
+  lazy val ObjectTypeDefinition: Parser[Ast.ObjectTypeDefinition] =
     for {
```
johnynek, line 58: replace flatMap with mapN

modules/core/src/main/scala/parser.scala
```diff
@@ -3,96 +3,98 @@
...
+    ObjectTypeDefinition.backtrack |
+    InterfaceTypeDefinition.backtrack |
+    UnionTypeDefinition.backtrack |
+    EnumTypeDefinition.backtrack |
     InputObjectTypeDefinition
 
-  lazy val ScalarTypeDefinition =
+  lazy val ScalarTypeDefinition: Parser[Ast.ScalarTypeDefinition] =
     for {
-      desc   <- opt(Description)
```
johnynek, line 49: replace flatMap with mapN

modules/core/src/main/scala/parser.scala
```diff
@@ -3,96 +3,98 @@
...
     } yield Ast.RootOperationTypeDefinition(optpe, tpe)
 
-  lazy val TypeDefinition =
-    ScalarTypeDefinition.widen[Ast.TypeDefinition] |
-    ObjectTypeDefinition |
-    InterfaceTypeDefinition |
-    UnionTypeDefinition |
-    EnumTypeDefinition |
+  lazy val TypeDefinition: Parser[Ast.TypeDefinition] =
+    ScalarTypeDefinition.widen[Ast.TypeDefinition].backtrack |
```
johnynek, line 42: I'd work hard to get rid of these backtracks since they could easily hide exponentially bad performance and obscure where the parser went wrong.

modules/core/src/main/scala/parser.scala
```diff
@@ -3,96 +3,98 @@
...
+  lazy val SchemaDefinition: Parser[Ast.SchemaDefinition] =
     for {
       _        <- keyword("schema")
       dirs     <- opt(Directives)
       rootdefs <- braces(many(RootOperationTypeDefinition))
     } yield Ast.SchemaDefinition(rootdefs, dirs.getOrElse(Nil))
 
-  lazy val RootOperationTypeDefinition =
+  lazy val RootOperationTypeDefinition: Parser[Ast.RootOperationTypeDefinition] =
     for {
```
johnynek, line 35: replace flatMap with mapN

modules/core/src/main/scala/parser.scala
```diff
@@ -3,96 +3,98 @@
...
     ExecutableDefinition | TypeSystemDefinition // | TypeSystemExtension
 
-  lazy val TypeSystemDefinition =
-    SchemaDefinition | TypeDefinition | DirectiveDefinition
+  lazy val TypeSystemDefinition: Parser[Ast.TypeSystemDefinition] =
+    SchemaDefinition | TypeDefinition.backtrack | DirectiveDefinition
 
-  lazy val SchemaDefinition =
+  lazy val SchemaDefinition: Parser[Ast.SchemaDefinition] =
     for {
```
johnynek, line 28: replace flatMap with mapN

modules/core/src/main/scala/parser.scala
```diff
@@ -242,7 +244,7 @@ object GraphQLParser {
       sel  <- SelectionSet
     } yield Ast.Selection.InlineFragment(cond, dirs, sel)
 
-  lazy val FragmentDefinition =
+  lazy val FragmentDefinition: Parser[Ast.FragmentDefinition] =
     for {
```
johnynek, line 248: replace flatMap with mapN

modules/core/src/main/scala/parser.scala
```diff
@@ -210,29 +212,29 @@ object GraphQLParser {
...
   lazy val Argument: Parser[(Ast.Name, Ast.Value)] =
-    (Name <~ keyword(":")) ~ Value
+    (Name <* keyword(":")) ~ Value
 
   lazy val FragmentSpread: Parser[Ast.Selection.FragmentSpread] =
-    keyword("...") ~> (FragmentName, Directives).mapN(Ast.Selection.FragmentSpread.apply)
+    keyword("...") *> (FragmentName, Directives).mapN(Ast.Selection.FragmentSpread.apply)
 
   lazy val InlineFragment: Parser[Ast.Selection.InlineFragment] =
     for {
```
johnynek, line 240: replace flatMap with mapN

modules/core/src/main/scala/parser.scala
```diff
@@ -251,13 +253,43 @@ object GraphQLParser {
...
+    lazy val BooleanValue: Parser[Ast.Value.BooleanValue] =
+      token(booleanLiteral).map(Ast.Value.BooleanValue.apply)
+
+    lazy val ObjectValue: Parser[Ast.Value.ObjectValue] =
+      braces(many(ObjectField)).map(Ast.Value.ObjectValue.apply)
+
+    lazy val ObjectField: Parser[(Ast.Name, Ast.Value)] =
+      (Name <* keyword(":")) ~ rec
 
-  lazy val Value: Parser[Ast.Value] =
```
johnynek, line 260: parser is covariant isn't it? I don't think you need the widen.

modules/core/src/main/scala/parser.scala
```diff
@@ -158,10 +160,10 @@ object GraphQLParser {
...
 
-  lazy val DirectiveLocations =
-    opt(keyword("|")) ~> sepBy(DirectiveLocation, keyword("|"))
+  lazy val DirectiveLocations: Parser0[List[Ast.DirectiveLocation]] =
+    opt(keyword("|")) *> DirectiveLocation.repSep0(keyword("|"))
 
-  lazy val DirectiveLocation =
+  lazy val DirectiveLocation: Parser[Ast.DirectiveLocation] =
     keyword("QUERY")       .as(Ast.DirectiveLocation.QUERY).widen[Ast.DirectiveLocation]     |
     keyword("MUTATION")    .as(Ast.DirectiveLocation.MUTATION).widen                         |
```
johnynek, line 168: parser is covariant, I don't think you need widen.

modules/core/src/main/scala/parser.scala
```diff
@@ -210,29 +212,29 @@ object GraphQLParser {
 
   lazy val Selection: Parser[Ast.Selection] =
     Field.widen[Ast.Selection] |
-    FragmentSpread.widen       |
+    FragmentSpread.widen.backtrack       |
     InlineFragment.widen
```
johnynek, line 216: parser is covariant, no need for widen.

modules/core/src/main/scala/parser.scala
```diff
@@ -210,29 +212,29 @@ object GraphQLParser {
 
   lazy val Selection: Parser[Ast.Selection] =
     Field.widen[Ast.Selection] |
-    FragmentSpread.widen       |
+    FragmentSpread.widen.backtrack       |
```
johnynek, line 215: can we possibly get rid of backtrack or comment what kinds of expressions need it? Maybe as easy as reversing the order?

johnynek (commented):
made some comments, I hope they are helpful.

TimWSpence:
> made some comments, I hope they are helpful.

Thanks SO much @johnynek! They are massively helpful! I'm away for the next week but will go through them properly when I come back.

modules/core/src/main/scala/parser.scala
TimWSpence: Pre-existing but yes, will remove 👍

modules/core/src/main/scala/parser.scala
TimWSpence: Pre-existing but yes, will rmeove 👍

modules/core/src/main/scala/parser.scala
TimWSpence: The issue is that multiple things can be prefixed by a string description so if we parse that then we need to backtrack. I had some work in progress to restructure the AST so we parse the comment separately and then attach it to the selection so I'll carry on with that 👍

modules/core/src/main/scala/parser.scala
TimWSpence: I assume this is for performance reasons?

TimWSpence: @johnynek this actually also doesn't work as it infers `Parser0` as the result type (via `Tuple3SemigroupalOps`). I can work around it with `~` and nested tuples. Is there a nicer way?

tpolecat:
(That test failure is sporadic.)

tpolecat (dismissed):
This looks great, thanks! Also very cool of you to have a look @johnynek, mahalo!

milessabin:
This is awesome ... thanks!

TimWSpence:
> This is awesome ... thanks!

Pleasure! :)

tpolecat (approved):
This is great, thanks!

TimWSpence:
> This is great, thanks!

Pleasure! :)

milessabin:
Looks like the flaky subscription test again ... re-running ...

johnynek:
It would be interesting to see some benchmarks to see if the parsing speed changed in a noticeable way or not.

Please let me know if you hit any issues with cats-parse!

PR #163, opened 2021-08-11
Attempt to extract doobie Put.Advanced type name

milessabin:
Previously doobie Advanced Puts were mapped to a bogus SQL type name `"OTHER"`. Now we attempt to extract the underlying `schemaTypes` name.

This showed up in a scenario where we had a pair of lists at the same level in the structure with the elements of one having a Json field that was not present in the other. This resulted in Grackle synthesizing a column for a union as `(NULL: OTHER)` rather than
`(NULL: json)`.


TimWSpence (approved):
thank you! :)

PR #165, opened 2021-08-17
Use type from `schemaTypes` in case of an array type for sql type name

### LINKED ISSUE #164: `array` is not valid type for PostgreSQL
When Grackle makes "union all" and in at least one field there is array field, then invalid SQL is produced.

The SQL contains:
```
SELECT ... (NULL :: array) ...
```
`array` is not valid type, it should be `type[]` or `type array` (e.g. `VARCHAR[]` or `VARCHAR ARRAY`).

I created a test case for that:
https://github.com/gemini-hlsw/gsp-graphql/compare/main...rpiotrow:topic/array-type-in-union-all-fixes

rpiotrow:
Fixes #164

### DIFF
--- a/modules/doobie/src/main/scala/DoobieMapping.scala
+++ b/modules/doobie/src/main/scala/DoobieMapping.scala
@@ -7,11 +7,12 @@ package doobie
 import cats.Reducible
 import edu.gemini.grackle.sql._
 import cats.effect.Sync
-import _root_.doobie.{ Meta, Put, Fragment => DoobieFragment, Read, Transactor }
+import _root_.doobie.{ Meta, Put, Read, Transactor, Fragment => DoobieFragment }
 import _root_.doobie.enumerated.JdbcType._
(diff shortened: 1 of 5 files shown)


modules/doobie/src/main/scala/DoobieMapping.scala
```diff
@@ -115,7 +115,7 @@ abstract class DoobieMapping[F[_]: Sync](
           case TinyInt               => Some("TINYINT")
           case VarBinary             => Some("VARBINARY")
           case VarChar               => Some("VARCHAR")
-          case Other                 =>
+          case Array | Other         =>
             codec._1.put match {
               case adv: Put.Advanced[_] =>
                 Some(adv.schemaTypes.head)
```
tpolecat, line 121: :+1: this will yield a type like `_varchar`, which is equivalent to `varchar[]`

tpolecat (approved):
Looks great, thanks!

milessabin (approved):
LGTM!

PR #166, opened 2021-08-23
expose compileUntyped

tpolecat:
I'm writing a routing layer that needs to know whether a query is a subscription or not. In order to do this I can call `QueryParser.parseText` and look at the resulting `UntypedOperation`, but there's no public way to compile this intermediate result down to an `Operation`. This PR just exposes that operation.


milessabin (commented):
I'm mostly fine with this, but can't you do what you need to do as a compiler phase?

milessabin:
I'm still not quite seeing what it is you're trying to do ... could you write some pseudo code annotated with "this bit is impossible" flagged up and I'll see if I can come up with something. Or even add a test to this to give me a clearer idea of how you plan to use it.

PR #169, opened 2021-08-25
Parser benchmarks and performance

TimWSpence:
Add a basic benchmark for the parser.

Also use this to make the parser ~1000X faster 😂


tpolecat (approved):
Great, thanks!

PR #170, opened 2021-08-26
Add Count query algebra term

milessabin:
Adds a `Count` term to the query algebra which counts the top-level elements of the result of it's nested `Query`.

In the doobie and skunk backends this is implemented in terms of a scalar subquery.

In the future we might generalise this to other aggregations.


tpolecat (approved):
:+1: this looks good! We need to document the query algebra at some point. Hopefully I'll have more time soon.

PR #172, opened 2021-09-01
Fix immediate self-joins

milessabin:
It turns out that up until now no one has attempted to represent a _directly_ recursive structure in Grackle ... all recursion has been indirect via one more more intermediate types.

A bug was revealed in the attempt to do this: we need to alias the child table of a join if it has already been seen, however prior to this PR, if the parent of the join is the same as the child, then both ends would be renamed, resulting in a join condition which would typically be empty.

This PR fixes that by ensuring that we only alias the child.


tpolecat (approved):
Looks great, thanks for all the tests! :+1:

PR #173, opened 2021-09-15
Fix aliasing of nested joins

milessabin:
Correctly alias all along the path in cases where a `SubObject` is reached via a nested join.

Also a couple of minor cleanups (whitespace, accidental use of `SqlUnionMapping` instead of `ObjectMapping` and an improved error report).


tpolecat (approved):
:+1:

PR #175, opened 2021-09-24
adding a new fieldMapping like CursorField but which returns Json

yasuba:


milessabin (approved):
LGTM!

It feels like there ought to be some way of unifying `CursorField`, `CursorJsonField` and `SqlJson` though ... they're basically three variations on the same theme.

PR #176, opened 2021-09-28
Big decimal not taken as a scalar

cquiroz:
This is not a code change but just a demonstration of an issue discussed in discord about scalar types

I think I tracked it to this line
https://github.com/gemini-hlsw/gsp-graphql/blob/main/modules/core/src/main/scala/schema.scala#L933

There I can see that `ScalarTypes` can only be the types defined in :
https://github.com/gemini-hlsw/gsp-graphql/blob/main/modules/core/src/main/scala/schema.scala#L459

### DIFF
--- a/modules/circe/src/main/scala/circemapping.scala
+++ b/modules/circe/src/main/scala/circemapping.scala
@@ -71,6 +71,7 @@ abstract class CirceMapping[F[_]: Monad] extends Mapping[F] {
         case e: EnumType       if focus.isString  =>
           if (focus.asString.map(e.hasValue).getOrElse(false)) focus.rightIor
           else mkErrorResult(s"Expected Enum ${e.name}, found ${focus.noSpaces}")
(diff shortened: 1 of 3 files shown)


milessabin:
Thanks!

tpolecat:
I added a case to handle this.

milessabin (approved):
LGTM!

modules/circe/src/main/scala/circemapping.scala
```diff
@@ -71,6 +71,7 @@ abstract class CirceMapping[F[_]: Monad] extends Mapping[F] {
         case e: EnumType       if focus.isString  =>
           if (focus.asString.map(e.hasValue).getOrElse(false)) focus.rightIor
           else mkErrorResult(s"Expected Enum ${e.name}, found ${focus.noSpaces}")
+        case _: ScalarType     if !focus.isObject => focus.rightIor // custom Scalar; any non-object type is fine
```
cquiroz, line 74: Huh, that simple :D

PR #178, opened 2021-10-29
Offset constructor

tpolecat:
This adds an `Offset` constructor to the query algebra, and handlers for `SqlMapping`.

This branch used to include the changes for cursor connections, now moved into #180

### DIFF
--- a/modules/doobie/src/test/scala/world/WorldData.scala
+++ b/modules/doobie/src/test/scala/world/WorldData.scala
@@ -62,7 +62,7 @@ trait WorldMapping[F[_]] extends WorldPostgresSchema[F] {
         cities(namePattern: String = "%"): [City!]
         city(id: Int): City
         country(code: String): Country
-        countries(limit: Int = -1, minPopulation: Int = 0, byPopulation: Boolean = false): [Country!]
+        countries(limit: Int = -1, offset: Int = 0, minPopulation: Int = 0, byPopulation: Boolean = false): [Country!]
         language(language: String): Language
         search(minPopulation: Int!, indepSince: Int!): [Country!]!
         search2(indep: Boolean!, limit: Int!): [Country!]!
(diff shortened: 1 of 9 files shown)


modules/doobie/src/test/scala/world/WorldData.scala
```diff
@@ -175,20 +175,21 @@ trait WorldMapping[F[_]] extends WorldPostgresSchema[F] {
...
         def limit(query: Query): Query =
           if (num < 1) query
           else Limit(num, query)
 
-        def order(query: Query): Query =
+        def order(query: Query): Query = {
           if (byPop) OrderBy(OrderSelections(List(OrderSelection(UniquePath[Int](List("population"))))), query)
-          else query
+          else       OrderBy(OrderSelections(List(OrderSelection(UniquePath[String](List("code"))))), query)
+        }
```
tpolecat, line 186: It's important to always have an explicit order if you're using OFFSET or LIMIT, otherwise the "paging" behavior won't really work. You get stuff back in arbitrary order.

modules/sql/src/main/scala/SqlMapping.scala
```diff
@@ -160,6 +160,7 @@ trait SqlMapping[F[_]] extends CirceMapping[F] with SqlModule[F] { self =>
           case u: UntypedNarrow => u.copy(child = loop(u.child, context))
           case n@Narrow(subtpe, _) => n.copy(child = loop(n.child, context.asType(subtpe)))
           case s: Skip => s.copy(child = loop(s.child, context))
+          case o: Offset => loop(o.child, context)
```
tpolecat, line 163: I think we can eliminate some other constructors here (like `Limit` and `OrderBy`) since they're handled by the database. We _must_ do it for `Offset` otherwise we will `drop` results we actually want to see!

tpolecat:
Ok this is ready for review.

tpolecat:
@milessabin ping, would be nice to get this in soon

modules/sql/src/main/scala/SqlMapping.scala
milessabin: I think we may need to retain `OrderBy` ... I think if we have multiple nested orderings there's no guarantee that the eventual order of rows internally is going to be preserved in the right way. But agreed that we _ought_ to be able to eliminate `Limit`.

milessabin (approved):
LGTM!

milessabin:
Rebased ...

PR #180, opened 2021-11-10
WIP: cursor connections

tpolecat:
This implements the [cursor connection pattern](https://relay.dev/graphql/connections.htm) for top-level SQL mappings by delegating to `SqlRoot` for the actual selection, then transforming the resulting cursor to conform with the connection schema. The interesting code and end-user doc are in `SqlConnectionMapping.scala`.


tpolecat:
Ok this is ready for review.

milessabin:
This looks great! My only concern is how does the usage of `ListCursor` interact with the existing usages of `List[Cursor]` eg. in `Cursor#asList` ... should the two be unified? Similar wrt `ScalarCursor` and `LeafCursor`.

Orthogonally, we really ought to work out how to factor out more of the doobie vs. skunk test duplication ... can we arrange it so that the _only_ thing that needs to be duplicated in the `TableDef`s?

tpolecat:
> This looks great! My only concern is how does the usage of ListCursor interact with the existing usages of List[Cursor] eg. in Cursor#asList ... should the two be unified? Similar wrt ScalarCursor and LeafCursor.

I think there's probably a lot of consolidation to be done around the construction and composition of cursors (there's also the `CursorBuilder` abstraction used by `GenericMapping`), but I'd prefer to take that on as a distinct issue.

> Orthogonally, we really ought to work out how to factor out more of the doobie vs. skunk test duplication ... can we arrange it so that the only thing that needs to be duplicated in the TableDefs?

:+1: Can't abstract over `TableDef` right now because it's defined independently in the doobie and Skunk mappings and is unknown to `SqlMapping`. Maybe we can hoist enough of it up to allow us to factor everything else out of the tests.

tpolecat:
> Can't abstract over TableDef right now

Actually I think I can. Trying something …

tpolecat:
Ok yeah, updated. I define a trait that defines all the `ColumnRef`s on it and then the subclasses can mix it into the backend-specific `TableDef`. Works fine. We can do this for most of the other tests (I think a few do things that aren't yet supported in Skunk).

kiendang:
Hi I'm interested in running a Relay compliant server with Grackle. May I know what's the state of this PR and is there anything I can do to help?

tpolecat:
We ended up doing something more general in #282 so this approach is abandoned (closing).

PR #182, opened 2021-11-12
Improved parser errors

TimWSpence:
Example query:
```
query {
  character(id: "1000" {
    name
}
```

Parser error:
```
Parse error at line 2 column 29
        character(id: "1000" {
                             ^
```

### DIFF
--- a/modules/core/src/test/scala/compiler/CompilerSpec.scala
+++ b/modules/core/src/test/scala/compiler/CompilerSpec.scala
@@ -360,7 +360,12 @@ final class CompilerSuite extends CatsSuite {
 
     val res = QueryParser.parseText(query)
 
-    assert(res == Ior.Left(mkOneError("Malformed query")))
+    val error =
+      """Parse error at line 2 column 29
+        |        character(id: "1000" {
+        |                             ^""".stripMargin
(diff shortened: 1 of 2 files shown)


modules/core/src/test/scala/compiler/CompilerSpec.scala
```diff
@@ -381,7 +386,12 @@ final class CompilerSuite extends CatsSuite {
 
     val res = QueryParser.parseText(query)
 
-    assert(res == Ior.Left(mkOneError("Malformed query")))
+    val error =
+      """Parse error at line 2 column 29
+        |        character(id: "1000" {
+        |                             ^""".stripMargin
```
milessabin, line 392: Oh! This test was _supposed_ to be for a missing closing `}` right at the end. I'd missed that that was being preempted by the missing closing `)` on the parameter block.

TimWSpence: 😂 I had wondered why they were the same!

milessabin (approved):
LGTM!

tpolecat:
This is great, thanks!

PR #188, opened 2022-05-12
Reworked the SQL compiler

milessabin:
+ The SQL compiler has been completely reworked, including ...

+ Improved representation of all SQL constructs.

+ Much less reliance on repeated SQL query rewrites during compilation. Even though these were pure, repeated rewriting made the logic unnecessarily hard to follow.

+ Complete reworking of the handling of table/column aliases. This is now deferred until SQL queries are rendered as Fragments.

+ Many bugfixes and performance improvements to the handling of filter/order by/offset/limit constructs.

+ Added support for Postgres window functions which are used in the compilation of nested filter/order by/offset/limit constructs.

+ Filter/offset/limit constructs which are representable as SQL are now fully handled in the DB and stripped out of the GraphQL result construction query.

+ Many bugfixes and performance improvements to the handling of multi-table joins.

+ Improved compilation of GraphQL type conditions via a TypeCase construct.

+ Interface/union discriminators are now handled in the DB. There has been a change in the user-facing mapping API changing the representation of discriminators to a first class SqlDiscriminator type.

+ The SqlMonitor trait now takes a count of rows and columns fetched rather than passing the complete result set.

+ Added a profile project allowing queries to be run against a dockerized Postgres instance and profiled with Java Flight Recorder.

+ More efficient representation of fetched data.

+ Various micro-optimizations around GraphQL schemas, types and mappings.


yasuba (approved):

tpolecat:
Wow! I will have a look this morning. Thanks!

profile/src/main/scala/Bench.scala
```diff
@@ -0,0 +1,379 @@
...
+    for {
+      _   <- IO.println("Warmup ...")
+      _   <- runQuery
+      _   <- runQuery
+      _   <- runQuery
+      _   <- runQuery
+      _   <- runQuery
+      _   <- IO.sleep(5.seconds)
+      _   <- IO.println("Running query ...")
+      _   <- runQuery
```
tpolecat, line 277: `replicateA` ;-)

tpolecat (approved):
Big, but it looks great. Looking forward to trying it out!

PR #227, opened 2022-07-07
Type conditions may widen as well as narrow

milessabin:
This PR fixes the incorrect assumption that a type condition (as part of a fragment, inline or otherwise) will only ever _narrow_ the expected type. Whilst it's pretty much pointless to widen in an `inline` fragment, standalone fragments might very well operate on types which are wider than the concrete types they're applied to, eg. in a case where a collection of supertype fields are collected up as a fragment and applied in multiple subtype subtrees.

For instance, given the schema,
```graphql
type Query {
  user(id: ID!): User!
  profiles: [Profile!]!
}
type User implements Profile {
  id: String!
  name: String!
  profilePic: String!
  friends: [User!]!
  mutualFriends: [User!]!
  favourite: UserOrPage
}
type Page implements Profile {
  id: String!
  title: String!
  likers: [User!]!
}
union UserOrPage = User | Page
interface Profile {
  id: String!
}
```
The following query contains a use of a fragment with a type condition (`on Profile`) which is a supertype of the static type which it is applied to (`User`),

```graphql
query withFragments {
  user(id: 1) {
    friends {
      ...profileFields
    }
  }
}

fragment profileFields on Profile {
  id
}
```

Prior to this PR, queries of the above form would either assert, or incorrectly execute as if the fragment didn't apply.

### DIFF
--- a/modules/circe/src/test/scala/CirceSpec.scala
+++ b/modules/circe/src/test/scala/CirceSpec.scala
@@ -175,6 +175,39 @@ final class CirceSpec extends CatsSuite {
     assert(res == expected)
   }
 
+  test("supertype fragment") {
+    val query = """
+      query {
+        root {
+          object {
+            ... on Child {
+              id
+            }
+            aField
+          }
+        }
+      }
+    """
+
+    val expected = json"""
+      {
+        "data" : {
+          "root" : {
+            "object" : {
+              "id" : "obj",
+              "aField" : 27
+            }
+          }
+        }
+      }
+    """
+
+    val res = TestCirceMapping.compileAndRun(query)
+    //println(res)
+
+    assert(res == expected)
+  }
+
   test("introspection") {
     val query = """
       query {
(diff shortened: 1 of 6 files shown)


tpolecat (approved):
Looks good. My only comment is that `conformsWith` may be a better name than `narrowsTo` after this change.

milessabin:
That makes sense. The only thing that give me pause is that `narrowsTo` and `narrow` correspond to `Narrow` in the query algebra. So if we rename the methods then we probably ought to rename the query algebra term too (we'd probably end up with `conformsWith`, `conform` and `Conform`). That will show through into any client code which traverses, constructs or transforms query algebra terms, and that's quite a lot client code.

I'm not opposed to doing this, but it will have an impact. We could add an `object Narrow` with (deprecated?) `apply`/`unapply` methods which are implemented in terms of `Conform`, but that's a little clunky.

WDYT?

milessabin:
I'll try a different approach ...

milessabin:
Much improved fix: instead of changing the behaviour of `Narrow`, `narrowsTo` and `narrow` we just don't generate a `Narrow` term when the target type is a supertype of the current static.

Apart from anything else, eliminating unnecessary `Narrow` terms means that in those case we don't need to inspect discriminators in the DB backed models, which simplifies generated SQL and improves performance.

PR #233, opened 2022-07-13
Fix directive location parsing

betehess:
Make sure that parsers for locations in directive definitions will correctly allow for backtracking. This does not happen if a directive location happens to be a prefix of another one **and** if the corresponding parser is attempted before that other one.


milessabin (approved):
LGTM!

PR #234, opened 2022-07-13
Fix `sbt docs/previewSite` on Java 17

betehess:
`sbt docs/previewSite` on Java 17 currently fails with

```
[error] java.lang.RuntimeException: Error creating extended parser class: Could not determine whether class 'org.pegdown.ParserWithDirectives$$parboiled' has already been loaded
```

The underlying issue and fix are discussed in https://github.com/sirthias/parboiled/issues/175

### DIFF
--- a/.sbtopts
+++ b/.sbtopts
@@ -1,0 +1,1 @@
+-J--add-opens=java.base/java.lang=ALL-UNNAMED


armanbilge:
Hmm, unfortunately adding this flag causes problems using JDK 8 in CI.

@milessabin what do you think about migrating the docs to sbt-typelevel-site as proposed in https://github.com/gemini-hlsw/gsp-graphql/issues/194#issuecomment-1135920317?

https://typelevel.org/sbt-typelevel/site.html

betehess:
> Hmm, unfortunately adding this flag causes problems using JDK 8 in CI.

Makes sense... Happy to update the PR so that the workaround is just mentioned in the docs i.e. use `SBT_OPTS="--add-opens java.base/java.lang=ALL-UNNAMED" sbt docs/previewSite` on Java 17.

milessabin:
@armanbilge I'd be very happy to migrate the docs to sbt-typelevel-site.

milessabin:
@betehess I can't reproduce this with Java 20 ... do you still need Java 17?

betehess:
@milessabin the workaround is good enough for me! Closing this.

PR #240, opened 2022-07-25
add parens around binops

tpolecat:
Filter expression `And(x, OR(y, z))` compiles to SQL expression `x AND y OR z` which means `(x AND y) OR z`, which is incorrect. This fixes it by adding parens around every binop.

We could try to get rid of extra parens when possible but this seems very low-priority to me.


tpolecat:
Rats, need to fix some tests.

milessabin (approved):
LGTM!

PR #242, opened 2022-07-29
switch Like to take an optional term

tpolecat:
The SQL `LIKE` operator can be applied to both nullable and non-nullable columns, however, prior to this commit, Grackle's `Like` predicate was only applicable to non-nullable `Terms`.

This PR allows `Like` to be applied to both nullable and non-nullable `Terms`. For Scala 3, this is captured directly using a union type for the `Term` (`Term[String]|Term[Option[String]]`). For Scala 2 we need a slightly more elaborate encoding using an implicit type constraint to ensure correctness outside the `sql` module, and a weaker type signature (`Term[_]`) internally.

A new `SqlLike` test suite has been added, which illustrates a plausible scenario where this might be desirable (as opposed to simply typing the term as `Term[String]` in all cases). In this example, method `mkPredicate` constructs a filter which applies a `Like` where a pattern is provided, or applies a null test when no pattern is provided. This method would be quite awkward to write if `Like` could only take a `Term[String]` as an argument because there is no straightforward way of obtaining a `Term[String]` from a `Term[Option[String]]` or vice versa.

### DIFF
--- a/modules/doobie/src/test/scala/DoobieSuites.scala
+++ b/modules/doobie/src/test/scala/DoobieSuites.scala
@@ -60,6 +60,10 @@ final class JsonbSpec extends DoobieDatabaseSuite with SqlJsonbSpec {
   lazy val mapping = new DoobieTestMapping(xa) with SqlJsonbMapping[IO]
 }
 
+final class LikeSpec extends DoobieDatabaseSuite with SqlLikeSpec {
+  lazy val mapping = new DoobieTestMapping(xa) with SqlLikeMapping[IO]
+}
+
 final class MovieSpec extends DoobieDatabaseSuite with SqlMovieSpec {
   lazy val mapping =
     new DoobieTestMapping(xa) with SqlMovieMapping[IO] {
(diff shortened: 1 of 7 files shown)


modules/core/src/main/scala/predicate.scala
```diff
@@ -33,6 +33,16 @@ trait Term[T] extends Product with Serializable { // fun fact: making this covar
 
   def forall(f: Term[_] => Boolean): Boolean =
     f(this) && children.forall(_.forall(f))
+
+  def map[U](f: T => U): Term[U] =
```
milessabin, line 37: Unfortunately this isn't going to work at all. The SQL compiler depends on being able to pattern match on terms to translate query algebra predicates in SQL conditions and this map completely obscures the underlying operation. We need to find some other way of representing optionality which is accessible to the pattern matcher.

tpolecat:
@milessabin thanks for rewriting it for me! Looks good. :+1:

milessabin (approved):
LGTM!

PR #246, opened 2022-08-05
Reduce use of Lists and miscellaneous performance tweaks

milessabin:
+ Uses of `List` replaced with `Seq`, `Iterable` or `Iterator` where possible.
+ `Cursor#asList` now takes a `Factory` to avoid committing to `List`.
+ `Context` path is more efficiently represented.
+ Several internal repeated list traversals and transforms replaced by hidden imperative loops for performance.
+ `SqlColumn` now records its result path which is used as a key for column lookup during result construction, eliminating many repeated linear traversals of the mappings.
+ Key columns and leaf encoders are memoized during result construction.

This reduces execution time of the following benchmark query by about 20%,

```graphql
query {
  cities {
    name
    country {
      name
      cities {
        name
      }
    }
  }
}
```

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -26,7 +26,7 @@ val Scala3 = "3.1.3"
 ThisBuild / scalaVersion        := Scala2
 ThisBuild / crossScalaVersions  := Seq(Scala2, Scala3)
 
-ThisBuild / tlBaseVersion    := "0.3"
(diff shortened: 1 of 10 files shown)


yasuba (approved):
lgtm

milessabin:
Bumped base version to keep MiMa happy ...

PR #250, opened 2022-08-10
improve error message

tpolecat:
Added the path and type and a suggestion that a `LeafMapping` is probably missing in the case where there's no encoder available for a result value.

Before:

```
Not a leaf: 0
```

After:

```
Cannot encode value 0 at programs/plannedTime/pi/microseconds (of GraphQL type NonNegLong). Did you forget a LeafMapping?
```

### DIFF
--- a/modules/sql/src/main/scala/SqlMapping.scala
+++ b/modules/sql/src/main/scala/SqlMapping.scala
@@ -3064,7 +3064,9 @@ trait SqlMapping[F[_]] extends CirceMapping[F] with SqlModule[F] { self =>
     def isLeaf: Boolean = tpe.isLeaf
 
     def asLeaf: Result[Json] =
-      mapped.encoderForLeaf(tpe).map(enc => enc(focus).rightIor).getOrElse(mkErrorResult(s"Not a leaf: $focus"))
+      mapped.encoderForLeaf(tpe).map(enc => enc(focus).rightIor).getOrElse(mkErrorResult(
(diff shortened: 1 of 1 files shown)


modules/sql/src/main/scala/SqlMapping.scala
```diff
@@ -3064,7 +3064,9 @@ trait SqlMapping[F[_]] extends CirceMapping[F] with SqlModule[F] { self =>
     def isLeaf: Boolean = tpe.isLeaf
 
     def asLeaf: Result[Json] =
-      mapped.encoderForLeaf(tpe).map(enc => enc(focus).rightIor).getOrElse(mkErrorResult(s"Not a leaf: $focus"))
+      mapped.encoderForLeaf(tpe).map(enc => enc(focus).rightIor).getOrElse(mkErrorResult(
+        s"Cannot encode value $focus at ${context.path.reverse.mkString("/")} (of type of type ${context.tpe}). Did you forget a LeafMapping?".stripMargin.trim
```
swalker2m, line 3068: `of type of type`

swalker2m (commented):
Excellent, newbies like me would benefit greatly from this kind of thing.

modules/sql/src/main/scala/SqlMapping.scala
tpolecat: doh

tpolecat: Fixed (description above edited).

milessabin (approved):
LGTM!

PR #251, opened 2022-08-12
convenience methods

tpolecat:
We have been using these as extension methods for a while and they're very useful.


milessabin (approved):
LGTM!

PR #252, opened 2022-08-15
add Cursor.envR

tpolecat:
Adds `envR`, makes `env` protected (to prevent confusion between `c.env[A]` and `c.env.get[A]`)


milessabin (approved):
LGTM!

PR #254, opened 2022-08-23
Add preunique operation to Cursor to support nested Unique

milessabin:
Previously the logic for changing the expected type from `T` to `[T]` when traversing through a `Unique` node in the query algebra was only handled for root mappings. This PR adds a `preunique` operation to `Cursor` which does this and which is used by the interpreter when evaluating `Unique` nodes.

### DIFF
--- a/modules/core/src/main/scala/cursor.scala
+++ b/modules/core/src/main/scala/cursor.scala
@@ -64,6 +64,12 @@ trait Cursor {
    */
   def asLeaf: Result[Json]
 
+  /**
+    * Yield a `Cursor` which can be used to evaluate the antecedant of a `Unique`
+    * operation.
+    */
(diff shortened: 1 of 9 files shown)


swalker2m (approved):
Thanks again for looking into this!  Looks great.

modules/core/src/main/scala/cursor.scala
```diff
@@ -64,6 +64,8 @@ trait Cursor {
    */
   def asLeaf: Result[Json]
 
+  def preunique: Result[Cursor]
```
tpolecat, line 67: Can you add a comment here explaining the intent?

tpolecat (approved):
Looks good, comment would be helpful. :+1:

PR #258, opened 2022-08-28
Parse type/schema extensions

### LINKED ISSUE #21: Support GraphQL extensions
The GraphQL spec defines a simple extension mechanism for all types. This allows the definition of a new type by the addition of new elements of the appropriate sort to an existing type (eg. an addition field to an object type). This is currently not supported.

This can most likely be implemented as an elaboration/validation phase.

keirlawson:
Added support for parsing schema/type extensions. Closes #21


milessabin:
Do want this reviewed as is? Or are you planning to extend the type representation correspondingly as well?

keirlawson:
Not sure what you mean by type representation? `Schema` and its components?

At any rate if you're happy with me doing this incrementally then I'd be very happy to have this reviewed as-is and then follow up with a further PR to propagate the change elsewhere in the codebase.

milessabin (changes requested):
A couple of minor comments ... looking good to far .. thanks! :tada:

modules/core/src/main/scala/ast.scala
```diff
@@ -10,7 +10,12 @@ object Ast {
   sealed trait Definition
 
   sealed trait ExecutableDefinition extends Definition
-  sealed trait TypeSystemDefinition extends Definition
+  sealed trait TypeSystemDefinitionOrExtension extends Definition
+
+  sealed trait TypeSystemDefinition extends TypeSystemDefinitionOrExtension
+  sealed trait TypeSystemExtension extends TypeSystemDefinitionOrExtension
+
+  sealed trait TypeExtension extends TypeSystemExtension
```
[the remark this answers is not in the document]
keirlawson: I was copying the hierarchy in the graphql spec, I guess extensions don't define a type by themselves so are arguable not `TypeDefinition`s, and I think when it comes to `Schema` it will be handy to be able to treat extensions as a different kind of thing.  That said happy to change it up if desired :)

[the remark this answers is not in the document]
milessabin: No, you're absolutely right.

milessabin:
This all looks great so far ... thanks! :tada:

keirlawson:
Have now added support to the schema, one thing I wasn't sure about, for all the regular type definitions the directives seem not to be stored against the instances in the schema, not clear if this is deliberate or not?

keirlawson:
Also not clear how best to avoid bincomapt issues flagged by CI or whether this is not a concern pre-1.x?

milessabin:
> for all the regular type definitions the directives seem not to be stored against the instances in the schema, not clear if this is deliberate or not?

It's semi-deliberate. The spec only requires support for `@skip`, `@include` and `@deprecated` and these are all handled. Grackle supports those, and nothing else.

I guess that from the point of view of round-tripping from SDL this is a limitation, and I'd be very happy to see a more first-class representation of directives.

milessabin:
> Also not clear how best to avoid bincomapt issues flagged by CI or whether this is not a concern pre-1.x?

Yeah, that's fine ... you can make the MiMa error go away by bumping `tlBaseVersion` to 0.7.

keirlawson:
In terms of what to implement next I was thinking some sort of `completeTypes` method on schema that gives a view of the types post-extension, that could then be used by the query interpreter etc, does that seem right?

milessabin:
@keirlawson I must admit I was thinking it should be possible to implement the relevant operations directly over type extensions rather than synthesize non-extended types. See if you can extend `<:<` to cover extensions and maybe take it from there?

keirlawson:
Not sure I follow you, should `<:<` return true when an extension is compared to a type it extends? If so why is this useful?

milessabin:
Yes `extendedType <:< typeExtended` should return true. It's a subtype test ... it's used all over the place in the compiler and interpreter :smile:

keirlawson:
Ok, when you have time could you take a look at what I've done with `Schema` just to make sure we're still on the same page?

milessabin:
@keirlawson I'm a bit surprised that `TypeExtension` doesn't extend `NamedType` ... I think if it did it would simplify a lot of things. Are there any obstacles to doing that?

keirlawson:
> @keirlawson I'm a bit surprised that `TypeExtension` doesn't extend `NamedType` ... I think if it did it would simplify a lot of things. Are there any obstacles to doing that?

Not so far as I am aware, think I just misunderstood what you had intended :-)

Is the idea then that these extended types would also be listed under a Schema's `.types`? This would mean that this would return multiple versions of the same type right? IE both the unextended version and then additional versions for every extension? Or is the idea that we stick in some logic such that only the extended type gets exposed here? In which case what to do when there are multiple extensions to the same base type...

milessabin:
> what to do when there are multiple extensions to the same base type...

Hmm ... interesting. The spec (including the current working draft) is silent on the topic of multiple extensions of the same base type. It does appear to be permitted, and seems to be supported by at least some implementations. Presumably the intended semantics are that the effective type is the union of the base type and all the extensions.

> Is the idea then that these extended types would also be listed under a Schema's .types?

I'm having second thoughts now in light of the above.

We need to keep extensions as individual entities for serializing the schema, but I'm now not so sure that it makes sense to enumerate them under `types`.

I'm coming round to the idea that a lookup of a type by name should return a type with all extensions applied. And that makes defining `Type` operations over extensions seem less important. Which also makes it seem like maybe they shouldn't be subtypes of `Type` at all?

Apologies if I'm going back on everything I said above :cry:

keirlawson:
Ok that all makes sense, in terms of returning the types with extensions applied, is there a preference between doing that dynamically on request as a view, vs calculating these at parse time and retaining them in memory? Strikes me as the latter likely better for perf if that isn't premature optimisation...

keirlawson:
I suppose this is just ultimately just the difference between a `val` and a `def` :grin:

modules/core/src/main/scala/schema.scala
```diff
@@ -23,12 +23,57 @@ trait Schema {
...
 
-  /** The types defined by this `Schema`. */
-  def types: List[NamedType]
+  /** The types defined by this `Schema` prior to any extensions. */
+  def unExtendedTypes: List[NamedType]
+
+  /** The types defined by this `Schema` with any extensions applied. */
+  def types: List[NamedType] = {
+    unExtendedTypes.map { tpe =>
+      val exts = extensions.filter(_.extended.dealias =:= tpe)
```
keirlawson, line 32: is this `dealias` necessary? I wasn't sure if `=:=` would take care of it or if another level of dealiasing is required?

modules/core/src/main/scala/schema.scala
```diff
@@ -23,12 +23,57 @@ trait Schema {
...
+      if (exts.length == 0) {tpe} else { extendType(tpe, exts) }
+    }
+  }
+
+  private def extendType(extendee: NamedType, toApply: List[NamedTypeExtension]): NamedType = {
+    extendee match {
+        case TypeRef(_, _) => throw new RuntimeException("Invariant violated")
```
keirlawson, line 39: Not sure if this is the correct thing to do here?

modules/core/src/main/scala/schema.scala
```diff
@@ -23,12 +23,57 @@ trait Schema {
 
   def pos: SourcePos
 
-  /** The types defined by this `Schema`. */
-  def types: List[NamedType]
+  /** The types defined by this `Schema` prior to any extensions. */
+  def unExtendedTypes: List[NamedType]
+
+  /** The types defined by this `Schema` with any extensions applied. */
+  def types: List[NamedType] = {
```
keirlawson, line 30: With this implementation approach this has to be lazy rather than eager owing to the way schemas are constructed using mutable vars

milessabin:
Apologies, but I'm off on vacation for the next week ... I'll look at this as soon as I get back ...

modules/core/src/main/scala/schema.scala
milessabin: `=:=` on its own should be sufficient.

modules/core/src/main/scala/schema.scala
milessabin: Could we accommodate this case with `extendee.dealias match { ... }`?

modules/core/src/main/scala/schema.scala
milessabin: I think this will need to be a `lazy val` for performance reasons.

modules/core/src/main/scala/schema.scala
keirlawson: Compiler still complains about the match not being exhaustive

keirlawson:
So if you're happy with the overall approach now by my reckoning what's left is tests and some validation logic, anything I'm missing?

keirlawson:
I *think* this is now ready for review

keirlawson:
@milessabin sorry to bug you again but is there anything else I need to do here to get this in a mergable state?

milessabin:
@keirlawson I've rebased this onto the current state of the repo, and tweaked various things and added more tests.

I think it's ready to merge now ... do you want to give it a once over?

milessabin (approved):

keirlawson (commented):
lgtm!

milessabin:
Thanks ... and once again, my apologies for taking so long to get this over the line.

Issue #202, opened 2022-05-26
Cross-build for Scala.js


milessabin:
This won't work for the doobie/skunk modules, but it should be fine for the others, and it would be interesting to get other backends contributed from the node ecosystem.

armanbilge:
It _is_ doable for skunk, skunk runs on Node.js :)

As I mentioned in https://github.com/gemini-hlsw/gsp-graphql/issues/128#issuecomment-1135968136, this should be very easy for the actual sources. The difficulty right now are that your tests spin up Docker images directly, which will be difficult to port. If you're willing to make a change, here's two strategies:
- Setup the docker images externally, if possible. Skunk does this with Docker compose which is why it was easy to port its Node.js tests.
- Explore using a different client such as [whale-tail](https://github.com/davenverse/whale-tail) which cross-builds for JVM and Node.js.

alexcardell:
TestContainers does have support for setting up arbitrary containers docker-compose files. 

There's also a testcontainers-node. Perhaps it's feasible to write some Scala.js testcontainers bindings to use the node/java docker-compose API depending on context?

armanbilge:
@alexcardell thanks for chiming in. I mentioned there's already an existing pure Scala Docker client that cross-compiles for JVM and Node.js (and will cross-compile for Scala Native as well without additional work, once its dependencies are ported). So I think investing in that is the best long-term solution, rather trying to juggle different implementations for the different platforms (dealing npm is not much fun).

https://github.com/davenverse/whale-tail

armanbilge:
This would benefit from https://github.com/gemini-hlsw/gsp-graphql/issues/128#issuecomment-1230772007.

PR #263, opened 2022-09-01
Cross-build for Scala.js

jf-botto:
Solves #202.

Just made a draft PR to get some opinions. I still want to rearrange/reformat a few things so that there's as fewer changes possible.

I'm very confused as to why: sbt compile; test works. It seems to be running all the tests and they all pass. I don't understand why.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -51,8 +51,8 @@ lazy val commonSettings = Seq(
   //scalacOptions --= Seq("-Wunused:params", "-Wunused:imports", "-Wunused:patvars", "-Wdead-code", "-Wunused:locals", "-Wunused:privates", "-Wunused:implicits"),
   scalacOptions += "-Wconf:msg=unused value of type org.scalatest.Assertion:silent",
   libraryDependencies ++= Seq(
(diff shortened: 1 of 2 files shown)


armanbilge:
https://youforgotapercentagesignoracolon.com/

You need to change all the dependencies to use `%%`, not `%%%`. The results can be completely misleading otherwise (tooling issues...).

jbotto94:
> https://youforgotapercentagesignoracolon.com/
> 
> You need to change all the dependencies to use `%%%`, not `%%`. The results can be completely misleading otherwise (tooling issues...).

Thank you so much!
I think the first issue I'm getting would actually be solved by getting Typename to compile to Scala.js, which I've noticed you have done herehttps://github.com/tpolecat/typename/pull/9...

armanbilge:
@jbotto94 this is now unblocked if you are motivated to keep working on it :)

jbotto94:
Will see what I can do

milessabin:
I've just merged that typename update :smile:

armanbilge:
Thank you for the early investigation! This landed in https://github.com/gemini-hlsw/gsp-graphql/pull/409.

PR #277, opened 2022-09-20
internalize unique/list distinction for paths

tpolecat:
This collapses the distinction between `UniquePath` and `ListPath` by requiring paths to specify their originating types. One can construct a path thus:

```scala
Path.from(tpe) // an empty path rooted at tpe
```

And can extend a path via `/`:

```scala
path / "a" / "b" / "c" // path extended by three more elements
```

As a convenience one can construct and extend a path via the new `/` method on `Type`:

```scala
tpe / "a" / "b" // an empty path rooted a t, extended with "a" and then "b"
```

In order to turn a path into a term one can call `.asTerm[A]` for any type `A`. This conversion is also applied implicitly, so a `Path` can be used in a context where a `Term` is expected. For example:

```scala
Eql(tpe / "foo" / "bar", Const(42)) // implictly convert a Path to a Term[Int]
```

The root type and underlying schema will determine whether the resulting `Term` has list or unique semantics. This makes no difference for predicates that are eliminated by the database back-end, but the distinction remains important for in-memory mappings.

### DIFF
--- a/modules/core/src/main/scala/predicate.scala
+++ b/modules/core/src/main/scala/predicate.scala
@@ -45,6 +45,7 @@ trait TermLow {
     p.asTerm[List[A]]
 }
 
+/** A path starting from some root type. */
 case class Path private (root: Type, path: List[String]) {
 
   def asTerm[A]: Term[A] =
@@ -64,6 +65,7 @@ case class Path private (root: Type, path: List[String]) {
 
 object Path {
 
+  /** Construct an empty `Path` rooted at the given type. */
   def from(tpe: Type): Path =
     Path(tpe, Nil)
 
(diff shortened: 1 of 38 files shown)


modules/core/src/main/scala/predicate.scala
```diff
@@ -35,54 +35,64 @@ trait Term[T] extends Product with Serializable { // fun fact: making this covar
...
-    * Reifies a traversal from a Cursor to multiple, possibly nullable, values.
-    *
-    * Typically such a path would use used to identify the list of values of
-    * an attribute of the object elements of a list field of the object or
-    * interface at the focus of the cursor, to be tested via a predicate such
-    * as `In`.
-    */
-  case class ListPath[T](val path: List[String]) extends Term[List[T]] with Path {
-    def apply(c: Cursor): Result[List[T]] =
+sealed trait PathTerm[A] extends Term[A] {
```
milessabin, line 74: `PathTerm`'s type parameter is the only thing I'm not completely sure about, because it's unused every where and clutters up the case clauses later, (ie. `case _: PathTerm[_]`). Could we get rid of it somehow? Does `PathTerm` actually need to extend `Term` ... `ListPath` and `UniquePath` do, but couldn't `PathTerm` be a standalone mixin? Or maybe we could relate it to `Term` via a self-type? eg. `trait PathTerm { self: Term[_] => ... }`.

milessabin (approved):
Just the one comment about the ergonomics of the type parameter of `PathTerm`. Your call on whether you feel like tweaking it or not.

modules/core/src/main/scala/predicate.scala
tpolecat: :+1: done

PR #281, opened 2022-09-27
Add the context.path to the mapping error messages.

mvalle:
Add the `context.path` to the mapping error messages, so that it's easier to identify the field that's causing the error.

The error message is in no small part inspired by https://github.com/gemini-hlsw/gsp-graphql/pull/250


tpolecat (approved):
:+1: Thanks!

milessabin:
I'm fine with this for now, but this is leaking internal errors back to clients and will be changing when https://github.com/gemini-hlsw/gsp-graphql/issues/199 is addressed. Provisionally what's going to happen is that all errors which aren't client-directed will go to a distinct internal error channel in `Result`.

PR #282, opened 2022-09-27
Reworked roots and added support for result transforms

milessabin:
+ `RootMappings` have been removed. Roots are now ordinary fields of Query/Mutation/Subscription mappings.

+ Mappings have been split into an `XxxMappingLike` trait which carries the definitions and logic, and an abstract class `XxxMapping` with a constructor. This enables arbitrary mappings to be mixed together while retaining a simple interface for client extension.

+ As a replacement for` RootMapping`, `Mapping` now defines a single `RootCursor` type which is focussed on the Query/Mutation/Subscription root type. Mapping subtypes contribute their own field mapping types by overriding `mkCursorForField`, which is called from the field method of `Cursor` during result construction traversal.

+ The effect logic of `RootMapping` has been moved to a new `RootEffect` type. `RootEffects` can perform an initial effect prior to computing a root `Cursor` and root query. See `MutationSpec` for an example.

+ A consequence of the above is that the various `Mapping` subtypes share more capabilities, for instance, the circe and generic mappings now support hidden and computed fields, and now also now support effects via the `RootEffect` infrastructure.

+ Added `TransformCursor` to the query algebra to support query result manipulation, post query execution, but prior to result construction.

+ Added `listSize`, `isDefined` and `envContains` to `Cursor` in support of `TransformCursor`.

+ The SQL mapping now supports a synthetic root table defined via `RootDef` allowing top level non-structured fields. This allows top level aggregation-like queries to be supported, eg. a count of elements of a collection.

+ Multiple root queries for a single mapping will now be handled together with respect to the SQL mappings. This means that multiple root queries will now be compiled and executed as a single SQL UNION query rather than as multiple queries sequentially. Combined with the previous bullet this means we can now, for example, fetch both a collection of items and an aggregate over that collection (eg. its size) in a single SQL query (in this case the query would be an SQL UNION of the count with the items).

+ Embedded subquery columns are now supported.

+ `In` predicates with an empty match set are now compiled to false, and child SQL queries will be eliminated where appropriate. Where this results in an empty top level SQL query this is now handled correctly.

+ The `CursorFieldJson` field mapping has been moved to the circe module.

+ Leaf encoders are now memoized in `Mapping` and the logic is shared between all `Mapping` subtypes.

+ The `LeafCursor` implementation has been moved to Mapping and is reused across all Mapping subtypes.

+ Added tests illustrating three approaches to paging.

+ Updated the code samples in the tutorial.

### DIFF
--- a/modules/core/src/main/scala/query.scala
+++ b/modules/core/src/main/scala/query.scala
@@ -5,6 +5,7 @@ package edu.gemini.grackle
 
 import scala.annotation.tailrec
 
+import cats.implicits._
 import cats.kernel.Order
 
 import Cursor.Env
@@ -83,7 +84,7 @@ object Query {
   case class Wrap(name: String, child: Query) extends Query {
     def render = {
       val rchild = if(child == Empty) "" else s" { ${child.render} }"
-      s"$name$rchild"
+      s"[$name]$rchild"
     }
   }
 
@@ -177,7 +178,15 @@ object Query {
   /** Computes the number of top-level elements of `child` as field `name` */
   case class Count(name: String, child: Query) extends Query {
     def render = s"$name:count { ${child.render} }"
-   }
+  }
+
+  /**
+   * Uses the supplied function to compute a continuation `Cursor` from the
+   * current `Cursor`.
+   */
+  case class TransformCursor(f: Cursor => Result[Cursor], child: Query) extends Query {
+    def render = s"<transform-cursor ${child.render}>"
+  }
 
   /** A placeholder for a skipped node */
   case object Skipped extends Query {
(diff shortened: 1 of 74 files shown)


modules/sql/src/test/scala/SqlPaging1Mapping.scala
```diff
@@ -0,0 +1,168 @@
...
+          List(
+            SqlField("id", city.id, key = true, hidden = true),
+            SqlField("countrycode", city.countrycode, hidden = true),
+            SqlField("name", city.name)
+          )
+      )
+    )
+
+  def genValue(key: String)(c: Cursor): Result[Int] =
+    c.env[Int](key).toRightIor(mkOneError(s"Missing key '$key'"))
```
tpolecat, line 123: FYI there is `envR` which returns a `Result` with a nice error message if the key is missing.

modules/core/src/main/scala/query.scala
```diff
@@ -227,14 +236,106 @@ object Query {
...
+        case TransformCursor(_, child) => loop(child, alias)
+        case _                         => None
+      }
+    loop(query, None)
+  }
+
+  /**
+   * Tranform the children of `query` using the supplied function.
+   */
+  def mapFields(query: Query)(f: Query => Result[Query]): Result[Query] = {
```
tpolecat, line 326: Useful!

tpolecat (approved):
:+1: This is great, thank you! The three paging examples are very helpful.

modules/sql/src/test/scala/SqlPaging1Mapping.scala
milessabin: Hmm ... I used `envR` elsewhere ... I must have missed this one ... good catch ... I'll tweak it later.

PR #294, opened 2022-10-19
Embedding fixes

milessabin:
When subobjects are embedded in an enclosing `SELECT` rather than via a join we have to ensure that all the corresponding context paths align. This has been done by wrapping the nested columns with `EmbeddedColumn` which preserves the embedded path of the underlying column whilst simultaneously matching the path of the enclosing select.

However a number of cases where this needed to be done (references in `WHERE` and `ORDER BY` clauses) were missed and also the generation of columns for parent constraints didn't take embedding into account.


tpolecat (approved):
🎉  This seems to resolve the problem. My real code is working now. Thanks!

tpolecat:
Hm on closer inspection when I run root-level paged queries I now get failures at `SqlMapping.scala:2798` because `extraCols` is non-empty ... it contains the bogus root-level column that's required to prevent a `No key columns` error.

PR #295, opened 2022-10-19
Aliasing Bug

tpolecat:
Here is a testcase that causes Grackle to produce invalid SQL due to an issue with column aliasing. It's not the exact problem I was seeing in my code but hopefully it has the same root cause. It's in the style of `SqlPaging2Mapping` but does away with the `hasNext` bit, which isn't relevant to the error.

The query in the test ends up generating the following:

```sql
SELECT
  t_observation_nested.c_observation_id,
  t_observation_nested.c_program_id,
  t_program.c_program_id AS c_program_id_alias_0,
  t_program.c_program_id AS c_program_id_alias_1
FROM
  t_program
  LEFT JOIN LATERAL (
    SELECT
      t_observation.c_program_id,
      (t_observation.c_observation_id COLLATE "C")
    FROM
      t_observation
    WHERE
      (
        (
          t_program.c_program_id_alias_0 = t_observation.c_program_id 
                    -------------------- 🔥 this isn't a column
        )
      )
    ORDER BY
      (t_observation.c_observation_id COLLATE "C") NULLS LAST
  ) AS t_observation_nested ON (
    t_observation_nested.c_program_id = t_program.c_program_id
  )
WHERE
  ((t_program.c_program_id = $ 1))
```

Note that the parent and child tables both use the name `c_program_id`, which forces the aliasing and may be the unusual circumstance that's exposing the issue.

### DIFF
--- a/modules/skunk/src/test/scala/SkunkSuites.scala
+++ b/modules/skunk/src/test/scala/SkunkSuites.scala
@@ -14,6 +14,7 @@ import edu.gemini.grackle.sql.SqlStatsMonitor
 
 import edu.gemini.grackle.sql.test._
 import edu.gemini.grackle.Mapping
+import skunk.AppliedFragment
 
 final class ArrayJoinSpec extends SkunkDatabaseSuite with SqlArrayJoinSpec {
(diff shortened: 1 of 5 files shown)


milessabin:
I'm fairly sure this was fixed in #294.

PR #297, opened 2022-10-23
Interface validation rework

keirlawson:
I've been reworking Grackle's interface validation code for schemas to more closely follow the algorithm outlined in the latest graphql spec and to cover some cases that the previous implementation did not (namely the former implementation did not deal with the case where a field is a subtype of that defined by the interface).  From this work a couple of design questions have arise that I'd be keen to get feedback on:

* I'm finding a need to dealias type aliases and invoke subtyping logic, both of these are already sipported by the `Schema` type however presently the validation is invoked before the schema is populated and so I'm wondering whether I should move where the validation happens, change the way schemas are constructed or re-implement this logic just for validation purposes?
* both objects and interfaces can implement interfaces but presently they lack a common supertype that exposes what interfaces they are attempting to implement, should I introduce one or adopt some other strategy?

### DIFF
--- a/modules/core/src/main/scala/schema.scala
+++ b/modules/core/src/main/scala/schema.scala
@@ -1056,6 +1056,112 @@ object SchemaValidator {
     errors.map(undefinedResults.addLeft(_)).getOrElse(undefinedResults)
   }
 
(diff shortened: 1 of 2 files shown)


milessabin:
I can't apologize enough for neglecting this PR.

Your observations were spot on and that's the approach I've taken in #426.

PR #307, opened 2022-11-25
Don't combine multiple top-level queries

milessabin:
0.8.x was a little over aggressive in combining the SQL queries generated for top-level query fields, potentially resulting in growth of the result set size quadratic in the number of top-level fields. This is sufficiently problematic in real world use, that it seems sensible to back this out pending a more intelligent combining rule.


tpolecat (approved):
Seems reasonable. 👍🏻

PR #308, opened 2022-11-29
Added support for composite keys

milessabin:
Mappings can now specify multiple parent/child join conditions for a single `Join`.


tpolecat (approved):
This looks great, thanks! 👍🏻

PR #310, opened 2022-12-07
WIP - add test to demonstrate issue with effectful, generic mapping

dantb:
This is a simplified version of a work problem, combining `RootEffect` and `GenericMapping`. I'm currently getting an error: `Mismatched query and cursor type in runValue: Query [Article]`

This makes sense conceptually, just not sure how to remedy it. Where am I going wrong @milessabin ?

### DIFF
--- a/modules/generic/src/test/scala/effectful.scala
+++ b/modules/generic/src/test/scala/effectful.scala
@@ -1,0 +1,77 @@
+// Copyright (c) 2016-2020 Association of Universities for Research in Astronomy, Inc. (AURA)
+// For license information see LICENSE or https://opensource.org/licenses/BSD-3-Clause
+
+package edu.gemini.grackle
+package generi
+
+import edu.gemini.grackle.generic.GenericMapping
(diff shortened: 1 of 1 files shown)


milessabin:
This should be fixed in https://github.com/gemini-hlsw/gsp-graphql/pull/313.

I've added a test [here](https://github.com/gemini-hlsw/gsp-graphql/pull/313/files#diff-47d9cced9158aafda6bc153d9f71cd7afd0f2e9fe6392f3ad02167ba6236d1c0), which looks fairly similar to yours in this PR.

dantb:
> This should be fixed in #313.
> 
> I've added a test [here](https://github.com/gemini-hlsw/gsp-graphql/pull/313/files#diff-47d9cced9158aafda6bc153d9f71cd7afd0f2e9fe6392f3ad02167ba6236d1c0), which looks fairly similar to yours in this PR.

Thanks @milessabin this looks great 😃

PR #312, opened 2022-12-07
add some convenience syntax

tpolecat:
This adds some convenience methods to `RootEffect` via syntax on `CirceMapping`, allowing computation of `F[Result[Json]]` and `F[Result[A]]` (given `Encoder[A]`) instead of computing a `CirceCursor`, which is usually entirely mechanical.


milessabin (approved):
LGTM!

PR #313, opened 2022-12-09
Fixed cursor construction for RootEffects

milessabin:
The only user-visible change is that the functions `RootEffects` are constructed with now take a `Path` rather than a root `Type`. Typically this path should be passed directly through to the relevant cursor construction call as was previously the case with the type, so, in many cases, user code which was previously correct will continue to compile and function as expected.

Fixes https://github.com/gemini-hlsw/gsp-graphql/pull/310.


tpolecat (approved):
This looks great. The change from `Type` to `Path` seems fine. 👍🏻

PR #314, opened 2022-12-10
Bumped Scala, SBT and dependency versions

milessabin:
Also eliminated a couple of non-local returns newly deprecated by Scala 3.2.x.


swalker2m (approved):
Looks great. :+1:

PR #321, opened 2022-12-21
Make CursorBuilder independent of mapping scope, add contramap

dantb:
On top of #320, will rebase if that goes in.

We've found this very handy, particularly for leaf nodes and wrappers around standard types (lists, options, etc).

The reason I haven't made it a Cats type class instance is that I wasn't sure if it's lawful? There are some [ContravariantMonoidalLaws](https://github.com/typelevel/cats/blob/main/laws/src/main/scala/cats/laws/ContravariantMonoidalLaws.scala). Maybe it should be called something else 🤷

### DIFF
--- a/modules/generic/src/test/scala/effects.scala
+++ b/modules/generic/src/test/scala/effects.scala
@@ -27,13 +27,14 @@ class GenericEffectMapping[F[_]: Sync](ref: SignallingRef[F, Int]) extends Gener
       }
     """
 
-  val QueryType = schema.ref("Query")
+  val QueryType  = schema.ref("Query")
   val StructType = schema.ref("Struct")
 
-  case class Struct(n: Int, s: String)
+  case class Wrapper(value: Int)
+  case class Struct(n: Wrapper, s: String)
   object Struct {
-    implicit val cursorBuilder: CursorBuilder[Struct] =
-      deriveObjectCursorBuilder[Struct](StructType)
+    implicit val wrappedCb: CursorBuilder[Wrapper]    = CursorBuilder[Int].contramap(_.value)
+    implicit val cursorBuilder: CursorBuilder[Struct] = deriveObjectCursorBuilder[Struct](StructType)
   }
 
   val typeMappings = List(
@@ -44,7 +45,7 @@ class GenericEffectMapping[F[_]: Sync](ref: SignallingRef[F, Int]) extends Gener
           // Compute a ValueCursor
           RootEffect.computeCursor("foo")((_, p, e) =>
             ref.update(_+1).as(
-              genericCursor(p, e, Struct(42, "hi"))
+              genericCursor(p, e, Struct(Wrapper(42), "hi"))
             )
           )
         )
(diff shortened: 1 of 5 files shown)


milessabin:
This and #320 are definite improvements ... thanks!

I'd be tempted to combine #320 and #321 ... there's no real need for them to be separate.

You'll need to bump `tlBaseVersion` to `0.11` to keep MiMa happy.

modules/generic/src/test/scala/effects.scala
```diff
@@ -76,15 +77,13 @@ final class GenericEffectSpec extends CatsSuite {
 
     val prg: IO[(Json, Int)] =
       for {
-        ref  <- SignallingRef[IO, Int](0)
-        map  =  new GenericEffectMapping(ref)
-        res  <- map.compileAndRun(query)
-        eff  <- ref.get
+        ref <- SignallingRef[IO, Int](0)
```
milessabin, line 80: I'd rather not have whitespace only changes.

modules/generic/src/test/scala/effects.scala
```diff
@@ -44,7 +45,7 @@ class GenericEffectMapping[F[_]: Sync](ref: SignallingRef[F, Int]) extends Gener
           // Compute a ValueCursor
           RootEffect.computeCursor("foo")((_, p, e) =>
             ref.update(_+1).as(
-              genericCursor(p, e, Struct(42, "hi"))
+              genericCursor(p, e, Struct(Wrapper(42), "hi"))
```
milessabin, line 48: I think a test for this would probably be better in the other test suite ... it doesn't have anything particularly to do with effects.

milessabin (changes requested):
Just those two niggles ... thanks! :smile:

modules/generic/src/test/scala/effects.scala
dantb: Ah yeah I was wondering whether you have a formatting tool? I applied scalafmt locally since I was too lazy to align things - is there any reason you don't already have it?

milessabin: I'm not a fan of code formatters. Maybe one day.

modules/generic/src/test/scala/effects.scala
```diff
@@ -76,15 +77,13 @@ final class GenericEffectSpec extends CatsSuite {
...
-        res  <- map.compileAndRun(query)
-        eff  <- ref.get
+        ref <- SignallingRef[IO, Int](0)
+        map = new GenericEffectMapping(ref)
+        res <- map.compileAndRun(query)
+        eff <- ref.get
       } yield (res, eff)
 
     val (res, eff) = prg.unsafeRunSync()
-    //println(res)
```
milessabin, line 86: Oh, also these should stay.

modules/generic/src/test/scala/effects.scala
```diff
@@ -113,15 +112,13 @@ final class GenericEffectSpec extends CatsSuite {
...
```
milessabin, line 123: These too.

dantb:
Poke @milessabin :)

milessabin (approved):
LGTM!

milessabin:
I'll merge this now and make a couple of changes in a follow-on PR,
1. Rename `CursorBuilder` to `GenericCursorBuilder`.
2. Drop the type alias of `CursorBuilder` in `GenericMappingLike`. Its only purpose is preserving backwards compatibility, and we don't really want to care about that until after a 1.0 release.

PR #337, opened 2023-01-25
Update to skunk 0.5.1

swalker2m:
A naive attempt to update skunk, mostly because of its natchez dependency which is conflicting elsewhere in another project that uses grackle, skunk, and natchez.  Any objections to this?


tpolecat (approved):

tpolecat:
Looks good, thanks.

PR #344, opened 2023-02-06
Handle failing discriminator correctly

tpolecat:
@milessabin this is the condition I was talking about earlier

Here I have created a subtype of `SqlInterfacesMapping` where the discriminator always fails, which should cause all type conditions to fail. Instead it causes all type conditions to succeed.

```graphql
query {
  entities {
    id
    entityType
    title
    ... on Film {
      rating
    }
    ... on Series {
      numberOfEpisodes
    }
  }
}
```

Should result in each result having only the common properties. Instead each result has both!

```json
{
  "data" : {
    "entities" : [
      {
        "id" : "4",
        "entityType" : "SERIES",
        "title" : "Series 1",
        "rating" : null,
        "numberOfEpisodes" : 5
      },
      {
        "id" : "5",
        "entityType" : "SERIES",
        "title" : "Series 2",
        "rating" : null,
        "numberOfEpisodes" : 6
      },
      {
        "id" : "2",
        "entityType" : "FILM",
        "title" : "Film 2",
        "rating" : "U",
        "numberOfEpisodes" : null
      },
      {
        "id" : "3",
        "entityType" : "FILM",
        "title" : "Film 3",
        "rating" : "15",
        "numberOfEpisodes" : null
      },
      ....
```

### DIFF
--- a/modules/doobie-pg/src/test/scala/DoobieSuites.scala
+++ b/modules/doobie-pg/src/test/scala/DoobieSuites.scala
@@ -73,6 +73,14 @@ final class InterfacesSpec extends DoobieDatabaseSuite with SqlInterfacesSpec {
     }
 }
 
+final class InterfacesSpec2 extends DoobieDatabaseSuite with SqlInterfacesSpec2 {
+  lazy val mapping =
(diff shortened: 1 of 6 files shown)


milessabin (approved):
LGTM!

PR #349, opened 2023-02-19
Fixes/refactorings of root effects to better support effect batching

milessabin:
+ Added `combineQueries` to `Mapping`.
+ Split `RootEffect` into an abstract `EffectMapping` type with `RootEffect` and `RootStream` subtypes. `RootEffect` represents a single shot effect whereas `RootStream` represents a stream for subscriptions.
+ Split root interpreter logic into subscription/`Stream` and non-subscription/non-`Stream` branches.
+ Moved some internal `QueryInterpreter` methods from yielding `Stream[F, T]` to yield `F[T]`.
+ Reworked `SqlComposedWorldMapping/Spec` to better illustrate effectful composition and batching.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -27,7 +27,7 @@ val Scala3 = "3.2.2"
 ThisBuild / scalaVersion        := Scala2
 ThisBuild / crossScalaVersions  := Seq(Scala2, Scala3)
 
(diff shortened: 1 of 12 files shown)


tpolecat (approved):
Cool, the `combineQueries` thing looks like a good general solution for batching. 👍🏻

PR #380, opened 2023-05-10
update to skunk rc

tpolecat:
We need to get our stuff on Skunk 0.6 ASAP so I'd like to merge this even though it's still an RC. I think we're probably the only people affected.


milessabin (approved):
LGTM!

PR #390, opened 2023-05-13
Add Ast.Document literals

dantb:
The error messages are pretty incomprehensible despite using the `Show` instance from cats parser. 

Shall I change all the parser tests to use the literal or is this alright as a quick sanity check?


milessabin:
I think this is just fine at is it ... thanks!

milessabin (approved):
LGTM!

PR #402, opened 2023-05-31
More informative error when implementing non-interface

milessabin:
Now reports that a non-interface type has been implemented rather than reporting that the named interface type does not exist.


modules/core/src/test/scala/schema/SchemaSpec.scala
```diff
@@ -310,6 +310,33 @@ final class SchemaSpec extends CatsSuite {
     }
   }
 
+  test("schema validation: implements non-interface") {
+    val schema =
```
cquiroz, line 314: Cool, thanks for the test

cquiroz (approved):
LGTM

PR #404, opened 2023-05-31
Removed most dependencies on SLF4J

milessabin:
It turns out we don't need an SLF4J dependency anywhere except the demo project. Even there it's satisfied transiently via http4s and doobie-hikari, but for hygiene's sake it's still listed explicitly.


armanbilge (approved):
Awesome, thanks!

PR #408, opened 2023-06-05
Move to MUnit and whale-tail

milessabin:
+ All tests reworked in terms of Munit.
+ All test (file)names normalised to "suite" from "spec".
+ Replaced test-containers with whale-tail to allow cross building for JS and native.


build.sbt
```diff
@@ -1,26 +1,29 @@
...
-val shapeless2Version           = "2.3.10"
-val shapeless3Version           = "3.1.0"
-val sourcePosVersion            = "1.1.0"
-val testContainersVersion       = "0.40.16"
-val typenameVersion             = "1.1.0"
+val catsVersion            = "2.9.0"
+val catsParseVersion       = "0.3.9"
+val catsEffectVersion      = "3.1.1"
+val circeVersion           = "0.14.5"
+val disciplineMunitVersion = "1.0.9"
```
armanbilge, line 8:
```suggestion
val disciplineMunitVersion = "2.0.0-M3"
```

modules/skunk/src/test/scala/SkunkDatabaseSuite.scala
```diff
@@ -19,15 +18,19 @@ import edu.gemini.grackle.sql.test._
 trait SkunkDatabaseSuite extends SqlDatabaseSuite {
 
   // Slow because each usage will open a new socket, but ok for now.
-  lazy val pool: Resource[IO, Session[IO]] =
+  lazy val pool: Resource[IO, Session[IO]] = {
```
armanbilge, line 21: We can replace this with a "resource fixture" that is shared for the entire suite in a follow-up PR.

modules/skunk/src/test/scala/SkunkDatabaseSuite.scala
```diff
@@ -19,15 +18,19 @@ import edu.gemini.grackle.sql.test._
...
+    import connInfo._
+
     Session.single[IO](
-      host     = container.containerIpAddress,
-      port     = container.mappedPort(POSTGRESQL_PORT),
-      user     = container.username,
-      password = Some(container.password),
-      database = container.databaseName,
-      // debug    = true,
+      host     = if (host == "0.0.0.0") "127.0.0.1" else host,
```
armanbilge, line 26: Note-to-self to look at this.

armanbilge (approved):
Admittedly only skimmed through an unfamiliar codebase but actually this seemed like a pretty straightforward migration, besides the whale tail stuff.

armanbilge:
After this merges I'll work on the JS and Native cross-builds. Thanks, these changes are a big help towards that!

tpolecat (approved):

build.sbt
milessabin: Done.

PR #409, opened 2023-06-06
Cross-build for JS+Native

armanbilge:
Almost there, just have to track down a couple things ... :)

### DIFF
--- a/modules/core/src/test/scala/GraphQLResponseTests.scala
+++ b/modules/core/src/test/scala/GraphQLResponseTests.scala
@@ -9,16 +9,16 @@ import munit.Assertions._
 
 object GraphQLResponseTests {
   def assertWeaklyEqual(x: Json, y: Json, strictPaths: List[List[String]] = Nil): Unit =
-    assert(weaklyEqual(x, y, strictPaths))
+    assert(weaklyEqual(clue(x), clue(y), strictPaths))
 
   def assertWeaklyEqualIO(x: IO[Json], y: Json, strictPaths: List[List[String]] = Nil): IO[Unit] =
-    x.map(x0 => assert(weaklyEqual(x0, y, strictPaths)))
+    x.map(x0 => assertWeaklyEqual(x0, y, strictPaths))
 
   def assertNoErrors(x: Json): Unit =
-    assert(noErrors(x))
+    assert(noErrors(clue(x)))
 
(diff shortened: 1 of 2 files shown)


armanbilge:
Darn.
- https://github.com/typelevel/skunk/pull/890

modules/core/src/test/scala/GraphQLResponseTests.scala
```diff
@@ -9,16 +9,16 @@ import munit.Assertions._
 
 object GraphQLResponseTests {
   def assertWeaklyEqual(x: Json, y: Json, strictPaths: List[List[String]] = Nil): Unit =
-    assert(weaklyEqual(x, y, strictPaths))
+    assert(weaklyEqual(clue(x), clue(y), strictPaths))
```
armanbilge, line 12: Anything wrapped in `clue` will be printed when the assertion fails, useful for debugging.

modules/sql/shared/src/test/scala/SqlWorldSuite.scala
```diff
@@ -1559,6 +1559,9 @@ trait SqlWorldSuite extends CatsEffectSuite {
       assertIO(resTotal, Some(expected))
   }
 
+  // JS treats floats as doubles
+  def isJS = Option(System.getProperty("java.vm.name")).contains("Scala.js")
```
armanbilge, line 1563: This was highly annoying but does the trick ...

modules/sql/js-jvm/src/test/scala/SqlDatabaseSuite.scala
```diff
@@ -26,7 +25,7 @@ trait SqlDatabaseSuite extends CatsEffectSuite {
   }
 
   def bindPath(path: String): String =
-    Path(".").absolute.parent.flatMap(_.parent).flatMap(_.parent).get.toString+"/"+path
+    buildinfo.BuildInfo.baseDirectory + "/" + path
```
milessabin, line 28: This is much better ... thanks!

modules/sql/shared/src/test/scala/SqlWorldSuite.scala
milessabin: Ouch, but :+1:

milessabin (approved):
This is awesome! Thanks so much :tada:

PR #410, opened 2023-06-06
Use circe-literal

armanbilge:
Instead of an inlined JSON literal.


milessabin:
@tpolecat I'm trying to remember why we had our own json literal ... was there no support for Scala 3 at some point?

armanbilge:
> was there no support for Scala 3 at some point?

yeah, it was added later

PR #415, opened 2023-06-07
Don't special-case `0.0.0.0` in skunk suites

armanbilge:
Opening to see what CI does ...


armanbilge:
Seems to be working fine in CI. @milessabin can you share more about the system where you needed this workaround?

milessabin:
I'll try JDKs which match CI.

armanbilge:
Hmm, ok, well now CI is failing 🤔

armanbilge:
I re-ran CI for the original commit and now that one also fails 🤷 sorry 😅

milessabin:
Simultaneously with you being able to trigger this in CI, I'm not currently able to reproduce the problem locally! :zany_face: 

This appears to be a very odd upstream (but which upstream?) bug, and I think we should just live with the workaround for now.

PR #426, opened 2023-06-15
Reworked schema validation

### LINKED ISSUE #68: Validate schema

milessabin:
+ Schema validation errors are now reported as failures rather than as warnings.
+ Schema validation is now primarily defined in terms of schema types rather than AST type definitions
+ All type references (field arguments and results, and implemented interfaces for both object and interface types, and union members for unions) are now checked.
+ Covariant subtyping between interfaces and implementations is now fully supported.
+ Subtype relation on Type now correctly handles non null and list type wrappers and transitive subtyping between interfaces.
+ Error messages have been made more consistent.

Fixes #68.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -30,7 +30,7 @@ val Scala3 = "3.3.0"
 ThisBuild / scalaVersion        := Scala2
 ThisBuild / crossScalaVersions  := Seq(Scala2, Scala3)
 
-ThisBuild / tlBaseVersion    := "0.13"
+ThisBuild / tlBaseVersion    := "0.14"
 ThisBuild / organization     := "edu.gemini"
 ThisBuild / organizationName := "Association of Universities for Research in Astronomy, Inc. (AURA)"
 ThisBuild / startYear        := Some(2019)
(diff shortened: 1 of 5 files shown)


tpolecat (approved):
Looks great, thanks! :+1:

PR #457, opened 2023-08-24
Disable BSP for JS/Native

armanbilge:
Disables the build-server-protocol for JS/Native. So Metals will only operate on the JVM projects.


tpolecat (approved):
👍🏻 this looks good to me and it fixes my issues with Metals being unusable. Any objections @milessabin?

milessabin:
Ahh ... damn I missed this PR despite the discussion on mastodon!

I've just opened #458 which switches to sbt-typelevel (from sbt-typelevel-ci) and uses `tlBspCrossProjectPlatforms`.

@armanbilge does my change do everything that this one does, and is there any reason for preferring one over the other?

armanbilge:
@milessabin thanks for doing that work! we should prefer your PR.

PR #458, opened 2023-08-24
Changes for latest sbt-typelevel

milessabin:
This switches from sbt-typelevel-ci to sbt-typelevel and does three things,
+ It sets the newly introduced `tlBspCrossProjectPlatforms` to only apply to the JVM platform.
+ It removes explicit dependencies to plugins which are included in the full sbt-typelevel.
+ It renames the root project to `root` to match the behaviour prior to the update of sbt-typelevel to 0.5.0.


armanbilge (approved):

milessabin:
@armanbilge thanks for that! I ran out of time last night! :tada:

PR #470, opened 2023-09-10
`extension` -> `extensions`

armanbilge:
I'm a little confused if this was an oversight in https://github.com/gemini-hlsw/gsp-graphql/pull/375 but according to the spec I believe this field should be rendered as `extensions`.

https://spec.graphql.org/October2021/#sec-Errors.Error-result-format


cquiroz (approved):
LGTM

PR #482, opened 2023-09-17
Use Ember in demo

armanbilge:
Ember is currently the blessed and best-supported backend.

Is there appetite to switch the demo from Doobie to Skunk?


milessabin (approved):
LGTM!

PR #483, opened 2023-09-17
Restore Scala.js and Native plugins

armanbilge:
These were removed by mistake in https://github.com/gemini-hlsw/gsp-graphql/pull/458. See:
- https://typelevel.org/sbt-typelevel/faq.html#how-do-i-manage-my-scala-js-and-scala-native-versions


milessabin:
Superseded by #486.

PR #493, opened 2023-10-18
Changed license to Apache 2

### LINKED ISSUE #196: Change license to Apache 2
Possibly controversial. But I think that as part of the move to the Typelevel organisation this would make sense.

milessabin:
Resolves #196.


armanbilge:
We need to change it here as well:
https://github.com/typelevel/grackle/blob/78c9e76fe9257db48e4606516057dc9617190b18/build.sbt#L36

milessabin:
Yup, I found that while porting to sbt-typelevel-site. Fix in progress.

PR #494, opened 2023-10-20
Moved docs from Paradox to sbt-typelevel-site

milessabin:
This is just moving the existing (local only!) Paradox based tutorial over to sbt-typelevel-site (ie. Laika and mdoc) with automated publication to https://typelevel.org/grackle. Much more to be done ...


modules/docs/src/main/scala/grackle/Output.scala
```diff
@@ -0,0 +1,39 @@
...
+import cats.effect.unsafe.implicits.global
+
+object Output {
+  def snip(path: String, tag: String): String = {
+    val header = "```scala"
+    val footer = "```"
+
+    val prg =
+      for {
+        txt <- IO.blocking(Files.readString(Path.of(path), StandardCharsets.UTF_8))
```
armanbilge, line 31: fwiw, at the moment there seems little value to doing this in `IO`, since we end up blocking anyway with `prg.unsafeRunSync()`.

docs/tutorial/intro.md
```diff
@@ -1,15 +1,6 @@
...
 * [In-memory model](in-memory-model.md)
-* [DB-backed model](db-backed-model.md)
-* [Composite model](composite-model.md)
-
-@@@
-
-## Table of Contents
-
-@@toc { depth=2 }
+* [DB Backed model](in-memory-model.md)
```
armanbilge, line 6: I think Laika has a comparable feature, might be useful.

https://typelevel.org/Laika/latest/07-reference/01-standard-directives.html#navigationtree

build.sbt
```diff
@@ -51,6 +52,9 @@ ThisBuild / githubWorkflowBuild     ~= { steps =>
 ThisBuild / githubWorkflowJavaVersions := Seq(JavaSpec.temurin("11"))
 ThisBuild / tlBspCrossProjectPlatforms := Set(JVMPlatform)
 
+ThisBuild / tlSitePublishBranch := Some("main")
+ThisBuild / tlSiteIsTypelevelProject.withRank(KeyRanks.Invisible) := Some(TypelevelProject.Affiliate)
```
armanbilge, line 56: Actually it's not an affiliate project, it's an org project: it is literally under the `typelevel` org on GitHub :)
```suggestion
```

build.sbt
```diff
@@ -28,12 +28,13 @@ val Scala2 = "2.13.12"
 val Scala3 = "3.3.1"
 ThisBuild / scalaVersion        := Scala2
 ThisBuild / crossScalaVersions  := Seq(Scala2, Scala3)
+ThisBuild / tlJdkRelease        := Some(11)
```
armanbilge, line 31: Is the intent to require JDK 11+ for Grackle?

build.sbt
```diff
@@ -99,6 +95,7 @@ lazy val modules: List[CompositeProject] = List(
   doobie,
   skunk,
   generic,
+  docs,
```
armanbilge, line 98: Typically I don't add the documentation site project to the root, but it's probably fine.

build.sbt
```diff
@@ -286,16 +280,12 @@ lazy val profile = project
   )
 
 lazy val docs = project
-  .in(file("docs"))
-  .enablePlugins(SitePreviewPlugin, ParadoxSitePlugin)
+  .in(file("modules/docs"))
+  .enablePlugins(NoPublishPlugin, AutomateHeaderPlugin)
+  .enablePlugins(TypelevelSitePlugin)
```
armanbilge, line 285: It already enables the no-publish plugin.
```suggestion
  .enablePlugins(TypelevelSitePlugin, AutomateHeaderPlugin)
```

armanbilge (commented):
Exciting!

docs/tutorial/intro.md
milessabin: For now I just want to get the site moved over. We can finesse things after that.

build.sbt
milessabin: I think yes.

modules/docs/src/main/scala/grackle/Output.scala
armanbilge: no, just an observation :)

armanbilge (approved):

PR #505, opened 2023-11-07
Rely on sbt-typelevel defaults for organization

milessabin:
`organizationName` should have changed with the move to Typelevel. Better to rely on the defaults in sbt-typelevel for this.


armanbilge (approved):
Good catch.

PR #506, opened 2023-11-07
Negative lookahead when parsing keywords

milessabin:
Fixes a bug in keyword parsing which became more obvious with the addition of type extensions, because this is the first time the parser has needed to recognize adjacent alphabetic keywords (eg. `extend type`).

Prior to this PR the parser would accept `extendtype` as equivalent to `extend type`. The [spec requires](https://spec.graphql.org/October2021/#Name) keywords to be parsed in the same way as names which have negative lookahead preventing them from being followed by a name continuation character. In practice this means that adjacent keywords must be separated by whitespace, comments or commas.

### DIFF
--- a/modules/core/src/main/scala/parser.scala
+++ b/modules/core/src/main/scala/parser.scala
@@ -26,7 +26,12 @@ import scala.util.matching.Regex
 
 object GraphQLParser {
 
-  def keyword(s: String) = token(string(s))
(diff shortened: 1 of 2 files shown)


milessabin:
Removed `.backtrack` as redundant on advice from @j-mie6.

milessabin:
Going with @j-mie6's review ...

Issue #183, opened 2021-11-30
add code coverage


tpolecat:
Let's set up code coverage to help find areas we haven't tested.

ahoy196:
Could you provide some guidance on what you would like to be the mechanism to set up the code coverage ? I'm a beginner and looking to make my first contribution to Grackle with this issue.

milessabin:
Hmm ... I'm not sure we still want to go down this route (in which case, my apologies for not closing this issue). The complication is that the existing Scala coverage tools are somewhat flaky and sensitive to Scala version and platform, and in the past have caused more administrative effort than justified by the benefits they've provided.

@tpolecat do you have a view on this?

armanbilge:
> and in the past have caused more administrative effort than justified by the benefits they've provided.

I think for now things seem okay. We've rolled out coverage to all the gemini-hlsw repos via sbt-lucuma. We also currently have coverage running in Skunk and http4s, specifically for JVM/2.13 which seems reliable enough.

FWIW, coverage is actually supported on JVM/JS/Native on Scala 2, and also Scala 3 but only JVM ...

milessabin:
@armanbilge can you point us to a project to model on? If we want coverage results available on codecov.io do we need typelevel account credentials?

armanbilge:
Here are two examples from Skunk and Cats Parse

https://github.com/typelevel/skunk/blob/1fad26f8838d689236f371909904d8e51f898c1d/build.sbt#L37-L47

https://github.com/typelevel/cats-parse/blob/79a414082fccb2e4db217712ae121d37494344a2/build.sbt#L32-L51

Pretty straightforward. No credentials necessary.

ahoy196:
Thanks @armanbilge and @milessabin !

@milessabin I will take a look at this ... 👍

ahoy196:
I posted a PR at #507 for review. Please take a look.

milessabin:
Fixed in #507 and #508. Thanks!

PR #507, opened 2023-11-10
Add coverage for Grackle

ahoy196:
This PR addresses #183 and adds support for code coverage to Grackle.

- Updated the build.sbt file 
- Updated the ci.yml file based on `githubWorkflowGenerate`
- Updated the README with a badge to show coverage based on `typelevel/grackle:main` - Note that for this to work and show the actual result, the uploading process to codecov should occur. This is all set up in this PR, but a repo token needs to be setup (I think in the environment).

Please review, and let me know if any further changes are required.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -50,6 +50,19 @@ ThisBuild / githubWorkflowBuild     ~= { steps =>
 ThisBuild / githubWorkflowJavaVersions := Seq(JavaSpec.temurin("11"))
 ThisBuild / tlBspCrossProjectPlatforms := Set(JVMPlatform)
 
+ThisBuild / githubWorkflowAddedJobs +=
+  WorkflowJob(
+    id = "coverage",
+    name = s"Generate coverage report (2.13 JVM only)",
+    scalas = List(Scala2),
+    steps = githubWorkflowJobSetup.value.toList ++
+      List(
+        WorkflowStep.Sbt(List("coverage", "coreJVM/test", "coverageReport")),
+        WorkflowStep.Use(UseRef.Public("codecov", "codecov-action", "v3"))
+      )
+  )
+
+
 ThisBuild / tlSitePublishBranch := Some("main")
 
 lazy val commonSettings = Seq(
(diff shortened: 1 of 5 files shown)


milessabin:
@ahoy196 brilliant! Many thanks for this :tada:

build.sbt
```diff
@@ -50,6 +50,19 @@ ThisBuild / githubWorkflowBuild     ~= { steps =>
 ThisBuild / githubWorkflowJavaVersions := Seq(JavaSpec.temurin("11"))
 ThisBuild / tlBspCrossProjectPlatforms := Set(JVMPlatform)
 
+ThisBuild / githubWorkflowAddedJobs +=
+  WorkflowJob(
+    id = "coverage",
+    name = s"Generate coverage report (2.13 JVM only)",
+    scalas = List(Scala2),
```
armanbilge, line 57: Let's copy the cats-parse setup here (we should port these changes to Skunk too). This way the CI workflow does not need to be re-generated every time there is a Scala version update.

```suggestion
    scalas = Nil,
    sbtStepPreamble = Nil,
```

armanbilge:
> do you know how to set up the repo token

I'm pretty sure it's not necessary. Codecov already accepted the results, and they are currently processing.

https://app.codecov.io/github/typelevel/grackle/commit/7786bbb81fd87de3fa866747359a6a997eecd833

milessabin (approved):
Nothing to add beyond @armanbilge's comments ... LGTM ... many thanks!

build.sbt
ahoy196: Good point. I've made the necessary recommended changes.

armanbilge (approved):

armanbilge:
Detailed coverage results are now available here: https://app.codecov.io/github/typelevel/grackle

milessabin:
That's great ... we probably want to exclude demo, benchmark and profile.

armanbilge:
Good point, we can do like http4s does.
https://github.com/http4s/http4s/blob/8744d6c4f62dfac1a2f55f7a928222325dc73eaa/build.sbt#L729

ahoy196:
I'll open a separate PR to exclude demo, benchmarks and profile.

PR #508, opened 2023-11-11
exclude demo, benchmarks and profile from coverage

ahoy196:
In the course of #507 @milessabin proposed that we exclude demo, benchmarks and profile from coverage for straightforward reasons. 

@armanbilge proposed a mechanism of doing so, which is the path followed in the current PR which addresses the same. This uses the `coverageEnabled` flag in sbt project settings.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -249,6 +249,7 @@ lazy val demo = project
   .settings(commonSettings)
   .settings(
     name := "grackle-demo",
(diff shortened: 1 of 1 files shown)


ahoy196:
I went ahead and excluded `docs` and `unidocs` from the coverage.

milessabin (approved):
LGTM!

PR #514, opened 2023-11-19
Change flyway dependency for demo module

rpiotrow:
There is a breaking change in Flyway 10.0.0, which introduced modularized database support. We need to use flyway module specific for PostgreSQL now, since core has no support for it.


milessabin (approved):
LGTM!

milessabin:
Well spotted! Thanks!

PR #516, opened 2023-11-22
Generated README.md using mdoc

milessabin:
The repo top-level `README.md` is now generated from `docs/index.md` and `header.md` using mdoc by manually running `repoDocs/mdoc`. In addition the repo top-level `CONTRIBUTING.md` file is now a symlink to `docs/CONTRIBUTING.md`. This means that the `docs` directory is now the source of truth for both the tutorial and the repo as viewed in github.


armanbilge (approved):
Very nice 👏

PR #519, opened 2023-11-24
Fixes for empty SQL sub-objects

milessabin:
The handling of empty SQL subobjects (previously represented by a context-free singleton `EmptySqlQuery`) was incorrect in several scenarios.

This showed up in particular in mixed mappings where non-SQL fields were nested in SQL objects. This manifested itself as unexpectedly empty arrays of subobjects, or possibly hit previously unimplemented methods of `EmptySqlQuery`. Both are fixed by ensuring that SQL ids are included when we, in effect, "join" a SQL with a non-SQL object. `EmptySqlQuery` is now a non-singleton and has a context and the unimplemented methods have been either implemented or removed.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -30,7 +30,7 @@ ThisBuild / scalaVersion        := Scala2
 ThisBuild / crossScalaVersions  := Seq(Scala2, Scala3)
 ThisBuild / tlJdkRelease        := Some(11)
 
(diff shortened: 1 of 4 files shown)


swalker2m (approved):
Thanks for working on this.  I tested it with my code that had been failing and it worked great.  :+1:

PR #521, opened 2023-11-27
Elaborate introspection in out of line fragments

### LINKED ISSUE #520: `__typename` in `fragment`
A co-worker uses a tool ([SGQLC](https://sgqlc.readthedocs.io/en/latest/)) to generate queries.  It seems to like inserting `__typename` wherever it can.  Once such place is in `fragment`s.   Unfortunately this may (?) not be working with Grackle.  For example this query is fine:

```
query {
  human(id: 1002) {
    friends {
      __typename
      ...CharacterFields
    }  
  }
}
fragment CharacterFields on Character {
  id
}
```

but this similar query fails with `No field '__typename' for type Character`:

```
query {
  human(id: 1002) {
    friends {
      ...CharacterFields
    }  
  }
}
fragment CharacterFields on Character {
  __typename
  id
}
```

milessabin:
Introspection fields in out of line fragments should be elaborated. Fixes #520.

Also removed some stray debugging code.

### DIFF
--- a/modules/core/src/main/scala/compiler.scala
+++ b/modules/core/src/main/scala/compiler.scala
@@ -234,7 +234,7 @@ class QueryCompiler(schema: Schema, phases: List[Phase]) {
       rootTpe <- op.rootTpe(schema)
       res     <- (
                    for {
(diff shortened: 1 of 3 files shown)


swalker2m (approved):
Excellent, thanks! :+1:

PR #525, opened 2023-12-01
Tweak to some collections usage

danicheg:
This brings some minor improvements to some collections usage. In some occurrences, it only slightly improves readability. Plus, brings some tiny micro-optimization of using `length` on Arrays instead of `size`. While such calls are legitimate, they require an additional implicit conversion to `SeqLike`. Anyway, this may not be that necessary — take it or leave it.


milessabin:
Umm ... wow, thanks. What was I smoking!?!

milessabin (approved):
LGTM!

PR #529, opened 2023-12-11
support for scalar argument without apostrophes

michludw:
GraphQL queries with custom scalar arguments (e.g.  `moviesLongerThan(duration: PT3H)`) are allowed by GraphQL and used to be supported in some previous grackle versions (before Elaborator rework).
This PR brings it back.

There is no need to support such inputs for variables since `untypedVars` are passed as Json and we can not pass the value without apostrophes.

I've decided to treat a custom scalar argument passed in enumeration-style (no apostrophes) as StringValue, so the GraphQL server always supports the input with and without apostrophes out of the box (it's enough to define support StringValue value handler).
Another option would be to keep the Enum value - this would allow to decide whether the enum-type value should be supported, but require the user to handle both StringValue and EnumValue values is so.

### DIFF
--- a/modules/core/src/main/scala/schema.scala
+++ b/modules/core/src/main/scala/schema.scala
@@ -1011,6 +1011,8 @@ object Value {
         value.success
       case (s: ScalarType, Some(value: BooleanValue)) if !s.isBuiltIn =>
         value.success
+      case (s: ScalarType, Some(enumValue: EnumValue)) if !s.isBuiltIn =>
+        StringValue(enumValue.name).success
 
       case (IDType, Some(value: IDValue)) =>
         value.success
--- a/modules/core/src/test/scala/compiler/ScalarsSuite.scala
+++ b/modules/core/src/test/scala/compiler/ScalarsSuite.scala
@@ -388,6 +388,39 @@ final class ScalarsSuite extends CatsEffectSuite {
     assertIO(res, expected)
   }
 
+  test("query with scalar argument without apostrophes") {
+    val query = """
+      query {
+        moviesLongerThan(duration: PT3H) {
+          title
+          duration
+        }
+      }
+    """
+
+    val expected = json"""
+      {
+        "data" : {
+          "moviesLongerThan" : [
+            {
+              "title" : "Celine et Julie Vont en Bateau",
+              "duration" : "PT3H25M"
+            },
+            {
+              "title" : "L'Amour fou",
+              "duration" : "PT4H12M"
+            }
+          ]
+        }
+      }
+    """
+
+    val res = MovieMapping.compileAndRun(query)
+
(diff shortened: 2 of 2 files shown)


modules/core/src/main/scala/schema.scala
```diff
@@ -1011,6 +1011,8 @@ object Value {
         value.success
       case (s: ScalarType, Some(value: BooleanValue)) if !s.isBuiltIn =>
         value.success
+      case (s: ScalarType, Some(enumValue: EnumValue)) if !s.isBuiltIn =>
+        StringValue(enumValue.name).success
```
michludw, line 1015: no need to handle EnumValue for scalar, always allows to skip the apostrophes
the other option would be
```
        enumValue.success
```
that allows to control whether to accept enum-style (with no apostrophes) values for scalars,
but also requires more handling later, e.g.:
```
object IntervalValue {
  def unapply(s: Value): Option[Duration] = s match {
    case StringValue(value) => Try(Duration.parse(value)).toOption
    case EnumValue(name) => Try(Duration.parse(name)).toOption
    case _ => none
  }
}
```
instead of
```
object IntervalValue {
  def unapply(s: StringValue): Option[Duration] =
    Try(Duration.parse(s.value)).toOption
  }
}
```

milessabin:
The spec doesn't actually have explicit wording that justifies this, however I've just checked with a GraphQL working group member, and it seems both that this is intended, even if not explicit, and also widely supported in many implementations.

This wouldn't be an enum, it would be using the literal form of the ID scalar type. Let me review the change I made that excluded it, and your proposed fix ... I'll try and get this, or something similar merged today.

milessabin:
> This wouldn't be an enum, it would be using the literal form of the ID scalar type

Update: no, this is wrong, an ID literal would still need quotes. So you're right that grackle was previously parsing the unquoted value as an enum value. I'll check again that this is acceptable.

modules/core/src/main/scala/schema.scala
milessabin: If this _is_ allowed by the spec, then I think you've made the right choice here.

milessabin:
This is the support we have,

> I don't see anything that explicitly forbids it.
> My intuition was that if not mentioned in coercion it would be in the validation section, but I don't see anything there forbidding it.
> I've not noticed this being used in the wild, but it does seem like it's not forbidden.

I would say that this is a grey area in the spec and basically "undefined and implementation dependent".

I think the likelihood is that if the spec _were_ tightened up it would _probably_ explicitly support the unquoted form, so I'll merge this PR now.

However I would encourage you to move away from the unquoted form in case the spec is revised in a way which excludes it, in which case Grackle would have to follow suit.

milessabin (approved):
LGTM!

PR #531, opened 2023-12-12
Fix for fragments with supertype refinements

milessabin:
This fixes several bugs in the SQL mappings handling of fragment spreads and inline fragments which _widen_ rather than narrow the context type.


camimora (approved):
LGTM! 👌

PR #534, opened 2023-12-14
fixing "TIMESTAMP WITH TIMEZONE" typo

michludw:
fixing a typo in the doobie timezoned time mapping;
there was `TIMEZONE`, it should be `TIME ZONE`

extending Coalesce test to cover this


milessabin:
Well spotted, and the fix looks good.

I'm not completely sure I understand why this is the right test to add to to exercise the fix though ... wouldn't it be better suited to `SqlMovieMapping/Suite` (which tests a range of custom scalars and PG data types)?

milessabin (approved):
Modulo the comments about the placing of the test ... LGTM!

michludw:
Hi, I was initially thinking about adding the test to the MovieSuite, but struggled to make it fail - it would require writing the query resulting in UNION (with date-time attribute skipped in one of the queries) if I get it right.
I will try to move the tests there...

milessabin:
That's good enough for me for now. Thanks again!

We probably need to revisit the structure of some of these tests.

PR #536, opened 2023-12-20
Add talks link

jatcwang:


milessabin:
Good call!

However, README.md is generated from docs/index.md, using mdoc. So the process is to make additions there; then in SBT run `repoDocs/mdoc` which will update README.md; then commit the result.

jatcwang:
Done :)

milessabin (approved):
LGTM!

PR #550, opened 2024-01-28
Fully implement field selection merging and collection rules

### LINKED ISSUE #22: Collect fragment fields correctly
See [5.3.2](https://graphql.github.io/graphql-spec/June2018/#sec-Field-Selection-Merging) and [6.3.2](https://graphql.github.io/graphql-spec/June2018/#sec-Field-Collection).

The spec defines behaviour when multiple applicable fragments request the same fields. This is not yet implemented.

milessabin:
See sections [5.3.2](https://spec.graphql.org/June2018/#sec-Field-Selection-Merging) and [6.3.2](https://spec.graphql.org/June2018/#sec-Field-Collection) of the GraphQL Specification.

Mergeability validation is done very early in compilation, immediately after variable and fragment use is validated.

As much field merging is done during compilation as possible. In some cases this will result in simpler executable queries than were previously generated which might slightly improve query execution time.

Field merging which depends on the runtime type of values has to be deferred to execution time, and potentially involves deep merging of `ProtoJson` result values. This could be expensive, and considerable effort has been taken to avoid it where possible and reduce its impact where not.

In almost all cases where this PR changes behaviour the previous behaviour was incorrect. The only changes in previously correct results are potential object field reorderings in Json results.

A handful of existing tests were violating the mergeability rules and have been corrected.

Fixes #22

### DIFF
--- a/modules/core/src/main/scala/compiler.scala
+++ b/modules/core/src/main/scala/compiler.scala
@@ -219,6 +219,7 @@ class QueryCompiler(parser: QueryParser, schema: Schema, phases: List[Phase]) {
     parser.parseText(text).flatMap { case (ops, frags) =>
       for {
         _    <- Result.fromProblems(validateVariablesAndFragments(ops, frags, reportUnused))
+        _    <- Result.fromProblems(validateFieldMergeability(ops, frags))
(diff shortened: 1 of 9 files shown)


tpolecat (approved):
This looks good. Thanks for the big new test suite!

PR #565, opened 2024-02-23
Cursors resolve fields internally before searching other mappings

milessabin:
Previously cursors attempt to resolve fields globally, potentially delegating to an alternative mapping, before resolving relative to the current cursor. In the case of the circe and generic mappings this could result in other mappings hijacking field resolution even when the currently focussed values could resolve the field successfully. Whilst this behaviour wasn't completely unmotivated, it runs counter to the way Grackle mappings are composed in practice.

This supersedes #460.


tpolecat (approved):
It would be good to have a test for this, but otherwise 👍🏻

milessabin:
Oops ... yes I forgot the test ... I'll adapt the one from your PR ...

PR #581, opened 2024-03-15
Added stack trace printing

phdoerfler:
Previously, many kinds of errors (e.g. a typo in a column name resulting in an exception from Postgres) would go unreported and the only clue that any kind of error occurred was a JSON decoding error in the playground.


milessabin:
That test failure looks like a (hopefully) transient issue with the docker test environment. I'm retrying :crossed_fingers:

milessabin (approved):
LGTM!

PR #592, opened 2024-04-09
Fixed fragment applicability to union types; no interfaces in unions

### LINKED ISSUE #590: Regression introduced in 0.18.1
Version 0.18.1 introduced a bug not present in 0.18.0.
The following query was working correctly in 0.18.0:

```
{
  a {
    ... AFields
  }
}

fragment XFields on X {
  x
}

fragment YFields on Y {
  y
}

fragment AFields on A {
  ... XFields
  ... YFields
}
```

With a schema

```
type Query {
  a: A!
}

union A = X | Y

type X {
  x: String!
}

type Y {
  y: String!
}
```

In 0.18.1 we get an error `Fragment 'AFields' is not compatible with type 'A'`

I'll try to find some time and write a test case

milessabin:
Fixes `fragmentApplies` so that it now accepts union types for a fragment type condition.

There are some lurking issues in the definition/use of `<:<` for unions which I don't want to address right now ... I'll create a separate issue to track that and pick them up later (#593)

In the process of exploring other test scenarios I realized that the GraphQL spec [explicitly excludes](https://spec.graphql.org/October2021/#sel-HAHdfFDABABlG3ib) interface types as being members of unions. I've modified the schema validator to check for that now, and fixed a couple of tests which became invalid as a result.

Fixes #590.

### DIFF
--- a/modules/core/src/main/scala/compiler.scala
+++ b/modules/core/src/main/scala/compiler.scala
@@ -1016,9 +1016,12 @@ object QueryCompiler {
      */
     def fragmentApplies(typeCond: NamedType, ctpe: NamedType): Boolean =
       (typeCond.dealias, ctpe.dealias) match {
-        case (_: InterfaceType, u: UnionType) => u.members.forall(_ <:< typeCond)
-        case (_, u: UnionType) => u.members.exists(typeCond <:< _)
-        case _ => typeCond <:< ctpe || ctpe <:< typeCond
+        case (u: UnionType, _) =>
+          u.members.exists(m => fragmentApplies(m, ctpe))
+        case (_, u: UnionType) =>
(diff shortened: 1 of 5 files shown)


milessabin:
@umazalakain I've sent you a collaborator invite ... please accept it if you like to review this :smile:

umazalakain (approved):
Agreed that it might be worth implementing a `fragmentApplies` that does not rely on the subtyping relation `<:<`. In particular,  https://spec.graphql.org/October2021/#sec-Fragment-spread-is-possible outlines a seemingly simple solution.

umazalakain:
alternative approach in https://github.com/typelevel/grackle/commit/735d231983820af768ec7b8b8a8c90b2a95d4111

PR #594, opened 2024-04-10
Fragments by possible subtype intersection

umazalakain:
As outlined in https://spec.graphql.org/October2021/#sec-Fragment-spread-is-possible


milessabin (approved):
LGTM!

Issue #606, opened 2024-04-29
Fragment changes number of elements in a list


phdoerfler:
It is currently possible to reduce the number of returned elements by adding a fragment to the query.
For details, see the two new tests added by the PR:
https://github.com/phdoerfler/grackle/commit/2693a18fa906bf7e43eea02a5b682a75f8156dfa.
This happened because of a mistake where I had used the wrong column in my doobie table. This wrong column would always be null in the case where the fragment applied. If you change `RichText!` into `RichText` in the schema, the number of objects would correctly remain three albeit with `null` as description for the affected element.
Changing the view in the .sql file to use `c.description` instead of the wrong `i.description` avoids this bug.

milessabin:
A perfect reproduction ... many thanks for that.

I'm really sure what I can do about this though. I think Grackle has to trust your view logic, and it has no independent way of determining how many elements there _ought_ to be in a result list.

I'm definitely open to ideas here, but for now I'm going to close this.

som-snytt:
if you close as "not planned", then it's easier to distinguish purple ("some useful work was done") from gray ("nothing to see here").

milessabin:
@som-snytt thanks for the tip!

PR #607, opened 2024-04-30
Added test for issue #606

phdoerfler:
This tests for the current behavior, i.e., the added tests both pass.


milessabin:
Many thanks for the reproduction.

See discussion in #606.

Issue #461, opened 2023-08-29
Improve prefix mapping mechanism


milessabin:
The current `PrefixMapping` can only specialise mappings based on a path prefix. Whilst this is fully general, it can be extremely verbose to specify mappings using only this mechanism.

There are at least two scenarios which are not currently handled well,
1. the mapping should be determined by a path from an ancestor type. If this type appears in multiple locations in the schema, then the current model requires multiple prefix mappings to encode it.
2. the mapping resolves a whole subtree in the result, eg. by fetching data as Json from a web service via a nested `EffectField`. If descendent types below the top level of that subtree have mappings elsewhere (eg. SQL mappings for a different path from the root) then these other mappings will supercede the effect mapping for those descendents. This can be worked around by the use of additional prefix mappings, but again that can be extremely cumbersome if the descendent types are used in many other parts of the schema.

We need a more sophisticated mapping mechanism which can handle these scenarios smoothly.

milessabin:
Fixed in #608.

Issue #230, opened 2022-07-11
False positives in validator re: unmapped types


tpolecat:
Some mappings (like Generic or Json) can traverse structures reflectively and don't need explicit mappings for contained types. The validator doesn't know this (can it?) and reports errors for unmapped types.

milessabin:
Yup, we're going to need to take a different approach to validation to make this work sensibly.

I think the right approach is to validate that all of a set of paths derivable from the schema are handled somehow. I think that's doable, at least to an reasonable approximation.

milessabin:
Fixed in #608.

Issue #229, opened 2022-07-11
Add nullity checking to mapping validator


tpolecat:
Nullity checking doesn't seem to be working, or isn't implemented. Follow up.

milessabin:
Fixed in WIP ...

milessabin:
Fixed in #608.

PR #608, opened 2024-04-30
Reworked mapping management and validation

### LINKED ISSUE #189: Mapping validator for DB mappings should check that all fields for a type map to columns of the same table
Grackle doesn't support splitting the fields of a type between multiple tables (other than as subobjects). This is not checked by validator.

Full example with test can be found here: https://github.com/rpiotrow/gsp-graphql/commit/f443b3d7785d9c0df990bf57310766e388b9ba5d

milessabin:
+ The collection of type mappings has moved to a new first class `TypeMappings` container type. There is an implicit conversion from `List[TypeMapping]` to `TypeMappings`, so this is a source compatible change (this will probably be deprecated in a later release). Most mapping validation logic has been moved to `TypeMappings`, apart from mapping-specific rules (eg. for `SqlMapping`) which are delegated to the relevant `Mapping` subtype at the granularity of validity for individual type and field mappings.

  Switching to the new API involves changing initialisations of the form,
  ```scala
  object SomeMapping extends ... {
    val schema = ...
    val typeMappings =
      List(
        ObjectMapping(
          tpe = SomeType,
          fieldMappings =
            List(
              SomeFieldMapping(...),
              ...
            )
        ),
        ...
      )
  }
  ```          
  to,
  ```scala
  object SomeMapping extends ... {
    val schema = ...
    val typeMappings =
      TypeMappings(
        ObjectMapping(SomeType)(
          SomeFieldMapping(...),
          ...
        ),
        ...
      )
  }
  ```

  While the current form is still supported, switching to the new form is encouraged to enable use of the new `MappingPredicates`. 
+ Added `MappingPredicate` as an extensible mechanism for matching mappings to paths. `TypeMatch` replaces the existing `Type` linked association, `PrefixedTypeMatch` replaces the `PrefixedMapping` wrapper, and `PathMatch` corresponds to the the `SwitchTypeMapping` used by Gemini/Aura.

  `PrefixedMapping` is still available in source compatible form, however the new form is encouraged. `PrefixedMappings` of the form,
  ```scala
  PrefixedMapping(
    tpe = SomeType
    mappings =
      List(
        List("some", "path") ->
          ObjectMapping(
            tpe = SomeType,
            fieldMappings =
              List(...)
            ),
      )
  )
  ```
  should be changed to,
  ```scala
  ObjectMapping(PrefixedTypeMatch(List("some", "path"), SomeType))(
    SomeFieldMapping(...),
    ...
  )
  ```
  Alternatively, using the new `PathMatch` predicate, which matches contexts with a path leading from some enclosing type,
  ```scala
  ObjectMapping(PathMatch(SomeParentType / "some" / "path"))(
    SomeFieldMapping(...),
    ...
  )
  ```  
  or more concisely,
  ```scala
  ObjectMapping(SomeParentType / "some" / "path")(
    SomeFieldMapping(...),
    ...
  )
  ```  

+ Mapping validation is now performed by default unless explicitly disabled. Validation is deferred until the mappings compiler is first referenced (to avoid init-order issues, object initialization errors and poor interactions of the latter with munit-cats-effect) and is performed exactly once. Applications which have at least one query unit test will trigger validation failures automatically at test time.

  While upgrading applications to this version, it might be desirable to temporarily suppress automatic validation. To do this, an existing list of type mappings can be wrapped in the `TypeMappings.unsafe` constructor,
  ```scala
  object SomeMapping extends ... {
    val schema = ...
    val typeMappings =
      TypeMappings.unsafe(
        List(
          ObjectMapping(
            tpe = SomeType,
            fieldMappings =
              List(
                SomeFieldMapping(...),
                ...
              )
          ),
          ...
        )
      )
  }
  ```
+ The validation algorithm has been reworked to start from the GraphQL schema and generate an exhaustive set of paths which are used to determine the relevant mappings to check against GraphQL and DB schema types. This supports both traversal through mappings which are partly implicit (eg. the Circe and Generic mappings) and also accommodate mappings which are guarded by `MappingPredicates`.
+ The base mapping validator now also reports unused (ie. not reachable via any valid path in the GraphQL schema) type and field mappings, and ambiguities where multiple mappings of the same specificity match the same path.
+ The SQL mapping validator checks that,
  + Nullability and Scala type are consistent between GraphQL and SQL.
  + Choice of `Leaf` or `ObjectMapping` is consistent with GraphQL leaf or non-leaf types.
  + GraphQL object types, interfaces, implementation and union members are nested within single a single DB table.
  + Associative fields are also keys.
  + Union field mappings must be hidden and leaf.
  + Embedded subobjects must be nested in their parent object DB table.
  + Joins must have at least one join condition.
  + Parallel joins must relate the same DB tables.
  + Serial joins must chain correctly.
+ Type mappings with non-trivial predicates are now indexed.
+ `LeafMapping` now supports `MappingPredicates` and are indexed along with `ObjectMappings`.
+ Built-in Scalar types now have explicit `LeafMappings` and can now be specialized for particular contexts via `MappingPredicates`.
+ Added a `subtree` Boolean attribute to `FieldMapping` to signal that a field value can represent structured result subtrees which are not explicitly mapped at all levels. This allows mapping validation to include the Circe and Generic mappings which implicitly map subtrees.
+ `ValueObjectMapping` has a new `on` constructor and the `withParent` initializer method, which was present on all field mappings has been removed and replaced by the `ValueFieldMapping` specific `unwrap`.

  The existing API is still supported, however moving to the new API is encouraged because it supports the use of `MappingPredicates`. To switch, mappings of the following form,
  ```scala
  ValueObjectMapping[Foo](
    tpe = FooType,
    fieldMappings =
      List(
        ValueField(...),
        ...
      )
  )
  ```
  should be changed to,
  ```scala
  ValueObjectMapping(FooType).on[Foo](
    ValueField(...),
    ...
  )
  ```
+ The `Path` type now supports narrowing GraphQL interface and union types along path. This allows the `PathMatch` mapping predicate to match mapping which are reachable from a base type along a path which passes through fields which have interface or union type fields. Narrowing is specified using the new `%` element as follows,
  ```scala
  SomeType / "interfaceField" % ImplementingType / "implementorSpecificField"
  ```
+ The `IntrospectionMapping` now uses the new mapping API.
+ All tests have been updated to validate under the new scheme.
+ The Doobie and Skunk test suites now have typed codecs which can capture the metadata needed for validation.
+ The tutorial and demo and profile projects have been updated to the new API.
+ Added GraphQL schema validation checks to ensure interfaces are non-cyclic.
+ Added `forUnderlyingNamed` convenience method to `Context`.
+ `ValidationFailure` has been split out to a separate file.
+ The `sql` classifier has been moved from `ValidationFailure` to a `SqlValidationFailure` subtype.
+ GraphQL types mentioned in `ValidationFailure` error messages are now correctly rendered using GraphQL syntax.
+ `PrimitiveMapping` is now redundant and all uses of it can be removed from application code.
+ `ComposedMapping` has been split out to a separate file.
+ Various signatures previously using `List` now use `Seq`.

Fixes #189, #229, #230, and #461.


tpolecat:
These all look like great improvements. It would be helpful if you could hold off on merging until Monday … I should be able to get our codebase tested by then. Will report back.

milessabin:
No problem.

I'm particularly keen to know if `PathMatch` is an adequate replacement for your `SwitchMapping`.

tpolecat (approved):
🎉  Lots of really good improvements here. Thanks for working through issues with me, it's all good now. :+1:

PR #634, opened 2024-06-11
Fix for polymorphic fields in SqlMapping

milessabin:
+ The resolution of polymorphic field mappings now takes the runtime context into account properly.
+ All narrowing operations on cursors yield a value in `Result` allowing mappings to report internal errors for failed discriminators.
+ In SQL mappings, failed discriminators now fail the entire query with an internal error.
+ The `TypeCase` extractor has been removed because it can't take the context into account.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -30,7 +30,7 @@ ThisBuild / scalaVersion        := Scala2
 ThisBuild / crossScalaVersions  := Seq(Scala2, Scala3)
 ThisBuild / tlJdkRelease        := Some(11)
 
-ThisBuild / tlBaseVersion    := "0.21"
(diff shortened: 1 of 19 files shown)


tpolecat:
Can you summarize what this change does? It looks like it's causing all of our discriminator predicates to fail.

tpolecat (approved):
:+1: this is working fine for us. The problem was on our end and was related to one of the new static methods on `ResultT`, for which we had a local syntax method with a different type (everything continued to typecheck after the update but some database operations were no longer being executed)

milessabin:
@tpolecat thanks for the update!

milessabin:
Rebased.

PR #646, opened 2024-08-16
preserve internal error

tpolecat:
We ran into an issue where an internal error in `SqlCursor.field` "expected exactly one row (or something like that)" was masked by `orElse` selecting the generic `mkCursorForField`, which then failed with "Unhandled mapping of type grackle.sql.SqlMappingLike$SqlField", which is misleading. This PR fixes the specific issue, but I think in general we may want a combinator like `r1 onFailure r2` that only selects `r2` in the case of logical failure, preserving `r1` as-is in case of `InternalError`.

### DIFF
--- a/modules/core/src/main/scala/result.scala
+++ b/modules/core/src/main/scala/result.scala
@@ -145,6 +145,12 @@ sealed trait Result[+T] {
       case _ => false
     }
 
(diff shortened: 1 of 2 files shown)


milessabin (approved):
LGTM!

PR #650, opened 2024-08-29
make effect elaborator respect predicated mappings

tpolecat:
This changes `EffectElaborator` to accept an arbitrary function that maps context+field to effect handler, allowing users to delegate to `typeMappings.fieldMapping` (which understands predicated mappings).

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -30,7 +30,7 @@ ThisBuild / scalaVersion        := Scala2
 ThisBuild / crossScalaVersions  := Seq(Scala2, Scala3)
 ThisBuild / tlJdkRelease        := Some(11)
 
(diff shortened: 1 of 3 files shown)


tpolecat:
Hm this is causing some failures on our end, looks like. Closing for now.

tpolecat:
Ok the failure was unrelated to this. It also may be the cause of our problems with #634 so I'll circle back to that.

tpolecat:
@milessabin we're blocked on this issue so it would be nice to get a release out if it looks ok to you.

milessabin (approved):
LGTM!

milessabin:
Yeah, that looks like it was an oversight part.

PR #662, opened 2024-10-05
Added support for Oracle via Doobie

milessabin:
Work in progress ...


tpolecat:
Oracle? Wow, my condolences.

milessabin:
@tpolecat the biggest job turned out to be rejigging the organisation of the various modules, and reworking the way docker is used in the tests/demo (the Oracle image takes way too long to startup and shutdown to run that cycle once for each suite).

I imagine you'll want to avert your gaze from the Oracle specifics, but I'd welcome your feedback on the reorg, the docker usage and also [this change](https://github.com/typelevel/grackle/pull/662/files#diff-9c151df62c5c3a1bd10416daa4a68548bbdbc026ff675f76ba68945707bbf774L43-L171) in the skunk module (currently a `Session` resource is `use`'d once per-`fetch`, which is probably harmless in the presence of a connection pool, but I think not ideal).

This will stay as WIP until the accompanying SQL Server module lands.

milessabin:
Superseded by #675.

PR #675, opened 2024-11-03
Modularised and added Oracle and SQL Server backends

milessabin:
+ Split `sql` module into `sql-core` and `sql-pg`, the latter containing Postgres specfics shared by `doobie-pg` and `skunk`.
+ Split off `doobie-core` from `doobie-pg` removing all Postgres specfics from the former.
+ Factored out all Postgres specifics from `SqlMapping`.
+ Added Oracle backend.
+ Added SQL Server backend.
+ Reduced the use of lateral subqueries to a minimum.
+ Simplified `SqlSelect` nesting logic.
+ Avoid nested conditions being captured by predicate subqueries.
+ ~~`SkunkMapping` now instantiated with a `Session` rather than a session `Resource` to avoid resource lifecycle on every query.~~ **update:** this turned out to be a Bad Idea and has been reverted. 
+ Replaced `whale-tail` with call outs to `docker-compose` allowing container to be shared by all suites across multiple test runs.

Running tests (eg. `rootJVM/test`) will automatically spin up the relevant containers, which will stay up and can be reused across multiple tests runs, significantly speeding up the test cycle. I recommend running `allUp` initially, to pull images and initialise databases, before running tests the first time. Note that Oracle in particular takes quite a long time to initialise, so expect `allUp` to take several minutes to complete the first time around.

There have been some changes in compilation of queries for Postgres, primarily a reduction in the usage of lateral subqueries (partly to simplify the implementation for SQL Server, which has a subtly different mechanism for achieving similar effects). These changes should preserve expected results, and reports of any changes in behaviour, including performance regressions, would be very much appreciated.


milessabin:
Agreed ... I'll add a note in the README.

milessabin:
@tpolecat Skunk changes made ...

@phdoerfler docs updated ...

milessabin:
:+1:'s from @tpolecat, @CamiloMorales and @phdoerfler ...

PR #679, opened 2024-11-15
Added checks for exit code 0

phdoerfler:
An attempt to escalate failures in the test setup so sbt can fail, skip the tests and present the user with a more meaningful error message.


milessabin (approved):
LGTM!

PR #692, opened 2025-01-10
Effect handler errors

swalker2m:
I have a use case in my application where a query accesses two fields which are implemented with `EffectField`s.  If both of these produce a `Result.Failure`, what should the caller expect?  In [my case](https://github.com/gemini-hlsw/lucuma-odb/blob/main/modules/service/src/test/scala/lucuma/odb/graphql/issue/shortcut/3660.scala#L15-L19) I sometimes get one error, sometimes two.  In this PR's test case, if you run it enough times you will see sometimes one field's failure is reported, sometimes the other.  I would like to always reliably get both errors or at least a deterministic result.

### DIFF
--- a/modules/circe/src/test/scala/CirceEffectHandlerErrorData.scala
+++ b/modules/circe/src/test/scala/CirceEffectHandlerErrorData.scala
@@ -1,0 +1,62 @@
+// Copyright (c) 2016-2023 Association of Universities for Research in Astronomy, Inc. (AURA)
+// Copyright (c) 2016-2023 Grackle Contributors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//   http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import cats.effect.Sync
+import cats.implicits._
+import fs2.concurrent.SignallingRef
+
+import grackle.Cursor
+import grackle.Query
+import grackle.Query.EffectHandler
+import grackle.Result
+import grackle.circe.CirceMapping
+import grackle.syntax._
+
+class TestCirceEffectHandlerErrorMapping[F[_]: Sync](ref: SignallingRef[F, Int]) extends CirceMapping[F] {
+  val schema =
+    schema"""
+      type Query {
+        n: Int!
+        s: String!
+      }
+    """
+
+  val QueryType = schema.ref("Query")
+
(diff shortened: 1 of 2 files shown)


milessabin:
Yes, I agree, you should be getting a consistent result here.

swalker2m:
Hoping to make some progress on this issue, I asked Claude Code to take a swing at it.  It produced this update at the end of `queryinterpreter.scala` in `completeAll` (replacing lines 768 to 813).  It appears to fix the issue.  First it puts the `EffectJson`s in a deterministic order, which are grouped as before but to the extent possible kept in the order they appear in the query.  Then it prevents short-circuiting the errors so that all the problems are reported.  Unfortunately I cannot push the update even to my own branch, but hopefully someone might be able to take a look?

```scala
    // Note: `gatherDeferred` prepends as it descends, so it yields deferred fields in reverse
    // document order; reversing once here recovers document order for everything below. We
    // group by `(mapping, handler)`, but derive the batch ordering (and the member ordering
    // within each batch) from document order rather than from `Map` iteration order, which is
    // non-deterministic and made the accumulated error order non-deterministic.
    val deferred = pjs.flatMap(gatherDeferred).reverse.asInstanceOf[List[EffectJson[F]]]
    val grouped  = deferred.groupMap(ej => (ej.mapping, ej.handler))(identity)
    val batchedEffects =
      deferred.map(ej => (ej.mapping, ej.handler)).distinct.fproduct(grouped)

    // Run every batch independently, then combine the per-batch `Result`s with an
    // accumulating combinator so that failures from *all* batches are preserved. Using
    // `ResultT.traverse` here would short-circuit on the first failed batch, and since the
    // batch order (from `groupMap(...).toList`) is non-deterministic, that produced a
    // non-deterministic subset of the errors.
    batchedEffects
      .traverse {
        case ((mapping, handler), batch) =>
          (for {
            pnext <-
              handler match {
                case None =>
                  ResultT(mapping.combineAndRun(batch.map(e => (e.query, e.cursor))))
                case Some(handler) =>
                  val queries = batch.map(e => (e.query, e.cursor))
                  for {
                    cs <- ResultT(handler.runEffects(queries))
                    conts <- ResultT(
                      queries
                        .traverse {
                          case (q, _) =>
                            Query
                              .extractChild(q)
                              .toResultOrError("Continuation query has the wrong shape")
                        }
                        .pure[F])
                    res <- ResultT(combineResults((conts, cs).parMapN {
                      case (query, cursor) =>
                        mapping.interpreter.runValue(query, cursor.tpe, cursor)
                    }).pure[F])
                  } yield res
              }
            next <- ResultT(completeAll[F](pnext))
          } yield batch.zip(next)).value
      }
      .map { results =>
        results.parSequence.map { batchedResults =>
          val subst = {
            val m = new java.util.IdentityHashMap[DeferredJson, Json]
            Monoid.combineAll(batchedResults).foreach { case (d, j) => m.put(d, j) }
            m.asScala
          }
          pjs.map(pj => scatterResults(pj, subst))
        }
      }
```

swalker2m:
See #854

PR #731, opened 2025-07-02
allow fetch to fail cleanly

tpolecat:
This changes `SqlModule.fetch` to yield a `Result`, which allows `fetch` to fail with cleanly (we are overriding `fetch` to fail on excessively large rowsets and prefer not to use an exception for this).

This also pulls out `rowDecoder` in `SkunkMapping` to allow access by overridden `fetch` implementations.

### DIFF
--- a/modules/doobie-core/src/main/scala/DoobieMapping.scala
+++ b/modules/doobie-core/src/main/scala/DoobieMapping.scala
@@ -147,7 +147,7 @@ trait DoobieMappingLike[F[_]] extends Mapping[F] with SqlMappingLike[F] {
       }
     }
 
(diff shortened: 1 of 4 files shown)


tpolecat:
Yes, the error will go back to the client with an explanation that their query was too complex. The alternative is to raise a special exception and catch it up top and turn it into a result, which is less appealing.

milessabin:
I'm still a little uncomfortable with this ... everything GraphQL related has been eliminated from the arguments and results of `fetch` so it seems odd that something GraphQL related should sneak out in the error channel. It feels like a layering violation.

I'm more than happy to merge the part that exposes the `rowDecoder`, and I'll merge the other part if you adamant, but it really feels to me that this sort of error should be returned in the `F` and turned into a GraphQL error somewhere where you actually have some GraphQL context.

PR #772, opened 2025-11-14
Add built-in `oneOf` directive

hugo-vrijswijk:
Adds a new directive built-in `@oneOf` that was added to the spec this september: https://spec.graphql.org/September2025/#sec--oneOf.

Specifies exactly one of the input fields must be defined and not null.

The spec mentions a special `OneOf Input Object`, but it seemed simpler to me to implement by adding the directive to the already-existing `InputObject`

Also adds two validations:
- Schema validation, to validate oneOf input definitions have no non-nullable fields
- Input validation, to validate inputs have exactly one field, and it is not null

### DIFF
--- a/modules/core/src/main/scala/introspection.scala
+++ b/modules/core/src/main/scala/introspection.scala
@@ -42,8 +42,6 @@ object Introspection {
           name: String
           description: String
 
-          isOneOf: Boolean
-
           # OBJECT and INTERFACE only
           fields(includeDeprecated: Boolean = false): [__Field!]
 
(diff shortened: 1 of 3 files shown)


milessabin:
This is great! Thanks so much for the contribution :smile:

I agree that using the existing `InputObject` type is the way to go.

I looks pretty good. I think you have tests covering most of the permutations [covered in the spec](https://spec.graphql.org/September2025/#sec-OneOf-Input-Objects) but I wonder if you could add something using variables, as in the examples shown in that section?

milessabin (approved):
LGTM!

PR #822, opened 2026-05-11
Update Native to v0.5 and Skunk to v1

hugo-vrijswijk:
And some other minor library updates.
Since Skunk 1 is only published to Native 0.5, there's no way to do these separately.


milessabin (approved):
LGTM!

milessabin:
Many thanks for that!

PR #832, opened 2026-06-04
Fix (-1, 0) numeric parsing

rpiaggio:
The `NumericLiteral` parser parsed the integer part as an `Int`, so for example for -0.99 the integer part -0 collapsed to 0, then the float was rebuilt from s"$a.$b" = "0.99".

The fix consists on capturing the integer part as a matched text instead of an int, thus keeping the negative sign.


milessabin (approved):

milessabin:
Many thanks!

PR #845, opened 2026-06-22
Fix accumulation of used variables in input objects

rpiaggio:
In `QueryCompiler.validateVariablesAndFragments`, only the last variable used in an input object was returned as used, it was not accumulating them in the recursion.


milessabin (approved):

milessabin:
Good catch! Many thanks.

PR #854, opened 2026-07-02
Accumulate effect handler errors deterministically

swalker2m:
## Problem

When a query selects two (or more) fields backed by `EffectField`s whose handlers each
return a `Result.Failure`, the response was non-deterministic: sometimes one field's error was
reported, sometimes the other, rarely both. Callers had no reliable way to see all of the
errors. This supersedes #692 (same change, re-opened from a fork after I lost push access to
the original branch).

Minimal reproduction is in `CirceEffectHandlerErrorSuite` / `...Data`: two fields `s` and `n`,
each on its own failing handler. Expected: both errors, in a stable order.

## Root cause

Effect results are completed in `QueryInterpreter.completeAll`. Deferred effects are grouped
into batches by `(mapping, handler)` and then merged. Two independent defects combined:

1. **Errors were dropped.** The batches were merged with `batchedEffects.traverse` running in
   `ResultT[F, *]`, whose `Applicative` delegates to `Result`'s *monadic* `ap`. That instance
   short-circuits on the first `Failure` and never accumulates, so only one batch's error
   survived. (Grackle already has an accumulating applicative — the `Parallel[Result]`
   instance — but it wasn't being used at this site.)

2. **Batch order was non-deterministic.** The batch list came from
   `groupMap(...).toList`, i.e. `Map` iteration (hash) order. Which batch was "first" — and
   therefore which single error survived defect #1 — varied from run to run.

Both effects always ran (the outer `F.map2` sequences them), so only the error *values* were
lost, not the side effects.

## Fix

In `completeAll`:

- **Accumulate across batches.** Run each batch to `F[Result[…]]` and combine the per-batch
  `Result`s with `parSequence`, which uses `Parallel[Result]` to accumulate all problems while
  correctly remaining a `Failure` when any batch fails. (`combineAllWithDefault` was rejected
  here because it downgrades an all-failed result to a `Warning` with empty substitutions,
  which breaks `scatterResults`.)

- **Deterministic, document-ordered batches.** `gatherDeferred` prepends as it descends, so it
  yields deferred fields in reverse document order. Reverse once, then group and take
  `distinct` keys, so both the batch order and the member order within each batch follow
  document order instead of `Map` iteration order.

## Tests

- `circe effect handler` — the original repro (two fields, two distinct handlers → two
  batches). Now deterministically reports both errors in document order.
- `circe shared effect handler` — new: two fields sharing *one* handler → a single batch whose
  `runEffects` receives both queries. Verifies both errors accumulate and appear in document
  order, and that both effects run.

## Note for reviewers: handler responsibility

The interpreter now preserves whatever a handler returns, but error accumulation *within* a
single batch is still the handler's job. A handler that combines its per-query results with
`sequence` (like the original test handler did) short-circuits and drops all but the first
error when two fields share it. The new shared-handler test uses `parSequence` to accumulate;
I confirmed that switching it to `sequence` reintroduces error loss. It may be worth
documenting this expectation for `EffectHandler` authors, or considering whether the framework
should defend against it.

### DIFF
--- a/modules/circe/src/test/scala/CirceEffectHandlerErrorData.scala
+++ b/modules/circe/src/test/scala/CirceEffectHandlerErrorData.scala
@@ -1,0 +1,62 @@
+// Copyright (c) 2016-2023 Association of Universities for Research in Astronomy, Inc. (AURA)
+// Copyright (c) 2016-2023 Grackle Contributors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//   http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import cats.effect.Sync
+import cats.implicits._
+import fs2.concurrent.SignallingRef
+
+import grackle.Cursor
+import grackle.Query
+import grackle.Query.EffectHandler
+import grackle.Result
+import grackle.circe.CirceMapping
+import grackle.syntax._
+
+class TestCirceEffectHandlerErrorMapping[F[_]: Sync](ref: SignallingRef[F, Int]) extends CirceMapping[F] {
+  val schema =
+    schema"""
+      type Query {
+        n: Int!
+        s: String!
+      }
+    """
+
+  val QueryType = schema.ref("Query")
+
+  case class TestEffectHandler[A](value: A) extends EffectHandler[F] {
+    def runEffects(queries: List[(Query, Cursor)]): F[Result[List[Cursor]]] =
+      queries.traverse { case (_, _) =>
+        ref.update(_ + 1).as(
+          Result.failure[Cursor](s"value: $value")
+        )
+      }.map(_.sequence)
+  }
+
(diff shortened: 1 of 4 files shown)


tpolecat:
This all makes sense to me, and thank for the writeup. I have run into issues with failure preservation a few times. Looks good to me but @milessabin needs to have a look.

milessabin:
Thanks for this.

My main issue is that it switches behaviour from fail-fast to accumulate all. I don't recall if the fail-fast behaviour was a deliberate decision or not, but it definitely seems legitimate and something we ought to preserve even if we want accumulate all as a option.

Where we _do_ want accumulate all, then a deterministic order is definitely desirable, and ideally document order. I think there's a simpler fix for this in `gatherDeferred`.

I'll noodle on it and push some changes to this PR later today or tomorrow.

milessabin:
I was mistaken when I said the status quo was fail fast ... all effects were executed, but only one error, picked non-deterministically, was reported. Clearly a bug ... thanks for persisting with this.

I've tweaked `gatherDeferred` so handlers are collected in document order with less `reverse` shenanigans, and made accumulate vs. (genuine) fail-fast a per-`Mapping` configurable option, with accumulation the default.

@swalker2m @tpolecat let me know if this works for you.

swalker2m:
Thanks Miles, this looks great. I don't think I can formally approve but :+1: from me. As you say, all effects were being executed before and they will still be executed by default after.  It's only the errors that weren't being accumulated.  Fail fast as a new option, with accumulate as the default (what we wanted in the task that triggered this PR) seems perfect.

milessabin (approved):

PR #864, opened 2026-07-12
Add SQLite backend

phdoerfler:
Adds SQLite as backend. SQLite does not support lateral joins, so `sql-core` had to be modified accordingly; that part is split out into #917, which this branch sits on top of. Merge #917 first, and this PR's diff shrinks to the new module and its test data.
This also opens the door for a H2 implementation which suffers from the same limitation as SQLite.

One small note related to the recent CSV-fication: `null-ordering`, `nullable-parent`, `qualified-names` and `union-order` have no `sqlite.sql`, so SQLite skips them.

Disclaimer: This work was partially done by an LLM but it was kept on a tight leash snd micromanaged by me.


milessabin (approved):
LGTM!

PR #866, opened 2026-07-13
Add H2 backend

phdoerfler:
This PR includes the commits of #917, which it needs for the lateral joins workaround. I would thus recommend merging that one first. It is independent of #864, the SQLite backend, though the two touch the same build and test-data files and whichever lands second will need a rebase.

Regarding the recent CSV-fication: `null-ordering`, `nullable-parent`, `qualified-names` and `union-order` have no `h2.sql`, so H2 skips them.

Disclaimer: This work was partially done by an LLM but it was kept on a tight leash snd micromanaged by me.


milessabin (approved):
LGTM!

PR #869, opened 2026-07-13
Add how-to guide: interfaces across multiple tables

phdoerfler:
Grackle's SQL module requires all implementors of an interface (or branches of a union) to be backed by a single table with a discriminator column. Users whose subtypes live in separate tables ask about this regularly, and the standing answer to use a VIEW has no public example anywhere in the repo or docs.

This adds a how-to page with a complete worked example: two source tables, the `UNION ALL` view with a synthesized discriminator and NULL-padded subtype columns, the (entirely ordinary) `SqlInterfaceMapping` over the view, a sample query/response, and the caveats that actually bite (cross-branch key uniqueness, `UNION ALL` vs `UNION`, predicate pushdown). It also introduces a "How-to Guides" docs section as a home for future recipes.

One visible side effect: listing the sections explicitly in `docs/directory.conf` moves Tutorial ahead of Contributing in the site nav (previously the unlisted tutorial sorted after it).

The code samples are plain fenced blocks rather than `mdoc`/`Output.snip` excerpts, to keep the change docs-only; they were verified end-to-end against PostgreSQL (the exact DDL, mapping, query, and response) before being copied into the page.

I was tempted to make this into a demo app or at least some sort of compile-able source code so that the CI can verify that the example code still compiles but I figured for now a simple page in the docs will be enough and I did not quite know where to best put that code.

### DIFF
--- a/docs/directory.conf
+++ b/docs/directory.conf
@@ -1,4 +1,6 @@
 laika.navigationOrder = [
   index.md
+  tutorial
+  howto
   CONTRIBUTING.md
 ]
--- a/docs/howto/directory.conf
+++ b/docs/howto/directory.conf
@@ -1,0 +1,4 @@
+laika.title = How-to Guides
+laika.navigationOrder = [
+  interfaces-across-tables.md
+]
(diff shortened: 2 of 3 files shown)


milessabin (approved):
Needs a rebase, but otherwise LGTM!

phdoerfler:
@milessabin Rebase done, ready to be merged 👍

PR #870, opened 2026-07-14
Fix NULL placement in MSSQL ORDER BY rendering

phdoerfler:
Here's another AI assisted bug fix. It's something that was uncovered while implementing the new backends and I have grilled the AI on this and it is convinced beyond a doubt that this is an issue in the current MSSQL backend. I think it checks out.

Here's what Claude has to say:

---

`DoobieMSSqlMapping.orderToFragment` renders the wrong NULL placement for one of the four `(ascending × nullsLast)` combinations: **descending with nulls first returns nulls last**.

SQL Server has no `NULLS FIRST`/`NULLS LAST` and sorts NULLs low (first ascending, last descending), so the two requests asking for the opposite placement are relocated with a leading sort key. The ascending correction is right, but the descending one inverts the key:

```scala
// nullsLast && ascending
" IS NULL THEN 1 ELSE 0 END ASC, "     // NULL→1 sorted ASC ⇒ nulls last ✅

// !nullsLast && !ascending
" IS NULL THEN 0 ELSE 1 END DESC, "    // NULL→0 sorted DESC ⇒ non-nulls first ⇒ nulls LAST ❌
```

In the second case the key sorts non-NULLs ahead of NULLs, which is precisely what SQL Server's unassisted default already does — so the correction is a no-op and the requested `NULLS FIRST` never happens. The fix uses the same key in both branches (`NULL → 1`) and lets the direction do the work: sorted `ASC` it pushes NULLs last, sorted `DESC` it pulls them first.

### Why no existing test caught it

Two things conspire to hide this:

* No shared fixture orders a nullable column whose data actually contains NULLs, so the shared suites can't observe where a dialect puts them.
* `stripCompiled` strips `Offset`/`Limit` (SQL is authoritative for those) but deliberately preserves `OrderBy`, so grackle re-sorts fetched rows in memory with correct, engine-independent NULL semantics. A dialect that renders NULL placement wrongly therefore still returns correct results for any query that fetches all rows.

SQL-side placement is only observable when it decides **which rows survive a LIMIT cut**. The new `NullOrderingSuite` is built around exactly that: it orders a NULL-bearing column with `limit: 2` and asserts the surviving ids, in order, across all four combinations.

With fixture rows `(1, 10) (2, NULL) (3, 5) (4, NULL)`, requesting *descending, nulls first* returned ids `[1, 3]` (the two non-null rows) where `[2, 4]` is correct. Against the unmodified dialect exactly one of the four tests fails — the other three, including `ascending, nulls last`, pass.

### Testing

* `NullOrderingSuite` fails on the current dialect (1 of 4) and passes with the fix (4/4).
* Full `doobiemssql/test`: 248/248 green.
* I ran the same suite against the other backends to see whether this was a wider problem: **Postgres and Oracle both pass** — they emit native `NULLS FIRST`/`NULLS LAST`, so they were never exposed. The bug is specific to dialects that have to emulate NULL placement.

If the suite looks generally useful, it would be a natural fit for `sql-core` so every backend gets it; happy to follow up with that separately rather than widen this PR.

### Note on CI

CI on this PR will fail at project load with `Your tlBaseVersion 0.28 is behind the latest tag 0.29.0` — that's #863, unrelated to this change, and currently affects every PR against `main`.


milessabin (changes requested):
This LGTM, though I think the test should be replicated across all the backends.

phdoerfler:
Very much agree. I actually have planned on moving the test to a place that would cover all backends and not just this one, but I wanted to keep that change in a separate PR because of all the other PRs adding new backends. Let me do that change right away...

phdoerfler:
Done. The test is now generalized into a shared `SqlNullOrderingMapping`/`SqlNullOrderingSuite` in `sql-core`.

milessabin (approved):
LGTM!

PR #871, opened 2026-07-14
Fold schema-qualified table names to bare identifiers in alias position

### LINKED ISSUE #342: SQL mapping does not work for recursive queries to DB tables with qualified names
### Summary

If DB tables with [qualified names](https://www.postgresql.org/docs/current/ddl-schemas.html) are used in a mapping, then recursive GraphQL queries do not work and fail with a cryptic error.

Note: checked for `DoobieMapping` only, not checked for `SkunkMapping`.

### How to reproduce

1. Apply this bug-reproducer patch https://github.com/gemini-hlsw/gsp-graphql/commit/ef0b21e6648fc0e8d2973c8aee1604ba73e426b5 to the current main branch.
2. Run Demo server (`reStart`)
3. Try this GraphQL request in the playground (or any other way of preference):
    ```graphql
    {
      country(code: "CAN") {
        name
        cities {
          name
          country {
            name
          }
        }
      }
    }
    ```

### Expected result

The query completes successfully. It does, actually, without the bug-reproducer patch applied. _Or_ it may work if a query does not contain recursive items, like `country / cities / country`.

### Actual result

The query that contains recursive items (like one shown above) fails to execute. In fact, it fails with 500 without any meaningful explanation.

By default, there's only a couple of messages in the logs:
```
demo [io-compute-2] INFO  o.h.s.m.Logger - service raised an error: class org.postgresql.util.PSQLException 
demo [io-compute-0] ERROR o.h.s.service-errors - Error servicing request: POST /world from ::1 
demo org.postgresql.util.PSQLException: ERROR: syntax error at or near "."
demo   Position: 315
demo 	at org.postgresql.core.v3.QueryExecutorImpl.receiveErrorResponse(QueryExecutorImpl.java:2674)
demo 	at org.postgresql.core.v3.QueryExecutorImpl.processResults(QueryExecutorImpl.java:2364)
demo 	at org.postgresql.core.v3.QueryExecutorImpl.execute(QueryExecutorImpl.java:354)
```
which unfortunately do not make a lot of sense.

However, the reproducer patch also adds `doobie.util.log.LogHandler.jdkLogHandler` to DB query builder. It helps to reveal a root-cause of the error:
```
demo[ERROR] Feb 02, 2023 3:47:59 PM doobie.util.log$LogHandler$ $anonfun$jdkLogHandler$1
demo[ERROR] SEVERE: Failed Statement Execution:
demo[ERROR] 
demo[ERROR]   SELECT  demo.city.id ,  demo.city.name ,  demo.country.code ,  demo.country.name AS name_alias_0 ,  demo.country_alias_1.code AS code_alias_2 ,  demo.country_alias_1.name AS name_alias_3  FROM  demo.country  LEFT JOIN  demo.city  ON  (demo.city.countrycode  =  demo.country.code )  INNER JOIN  demo.country AS demo.country_alias_1  ON  (demo.country_alias_1.code  =  demo.city.countrycode )   WHERE (( demo.country.code  =  ?) ) 
demo[ERROR] 
demo[ERROR]  arguments = [CAN]
demo[ERROR]    elapsed = 26 ms exec (failed)
demo[ERROR]    failure = ERROR: syntax error at or near "."
demo[ERROR]   Position: 315
demo[ERROR]           
```
Apparently, the compiled `SELECT` statement contains an alias (`demo.country_alias_1`) that is prepended with the DB schema name, which is not allowed and in fact is an SQL syntax violation.

### Work-around

In some cases it is possible to use the `before` strategy to work this issue around, e.g.:
```scala
val transactor: Transactor[F] = ???
val fixedTransactor =
  Transactor.before.modify(transactor, Update0("SET search_path TO demo").run *> _)
```
After that it becomes possible to use table names without the `demo` schema name in the mappings.

Not a real solution, of course.

phdoerfler:
Co-authored by an LLM. Reviewed by me. And now without further ado:

---

Fixes #342.

## Symptom

A mapping over a schema-qualified table name such as `public.country` produces invalid SQL as soon as the query shape needs an alias or a synthesized name for that table. Postgres reports `syntax error at or near "."`, because a dot is not legal in an identifier position:

```sql
INNER JOIN public.country AS public.country_alias_1
```

## Root cause

A table's SQL name has to serve two incompatible roles. As a *reference* it belongs in a FROM or JOIN clause fully qualified, dots intact. As an *identifier* — an alias, or a component of a synthesized subquery name — it must be bare. Grackle carried the name as a single opaque `String`, so every site that needed the identifier form silently got the reference form instead.

There is a second, subtler layer. In `SqlUnion.addFilterOrderByOffsetLimit` the name given to the wrapping subquery is not cosmetic: `TableExpr` identity is name equality, so naming the subquery after its parent table is precisely what keeps it `isSameOwner` with that table, and what lets columns still bound to the table resolve against the subquery. Folding the name there — treating it like the other synthesized names — repairs the alias but breaks the identity, and rendering then emits the raw dotted name against a FROM entry carrying the folded one:

```sql
FROM ( ... UNION ALL ... ) AS qualified_city
WHERE (qualified.city.id IS NOT NULL)
```

For an unqualified table the reference and identifier forms coincide, so neither layer is visible; both surface only once a schema qualifier is present.

## Fix

`TableName` becomes a structured `TableName(schema: Option[String], name: String)` with two derived accessors — `sqlRef` for the reference form and `identifier` for the bare form — and a `toString` that yields `sqlRef`, so the many sites that interpolate a table name into an error message keep rendering correctly with no change. `TableExpr` gains a corresponding abstract `identifier` member, and `ColumnRef.table` and `TableRef`'s field change from `String` to `TableName`. A call site that needs an identifier now reaches for `.identifier`; it can no longer reach for the raw name by accident, because the raw name is no longer a bare `String` sitting in scope. The fold itself is a single rule, `TableName.foldToIdentifier`.

Identifier positions are: the alias minted by `AliasState.tableDef`, the fallback in `AliasState.tableRef`, `syntheticName` (both the table and its joined children), the associative `_assoc` derived-table alias, the common table expression name minted by `subqueryToWithQuery`, and MSSQL's `encapsulateUnionBranch` `_encaps` sibling.

The union subquery is deliberately *not* in that list. It keeps its parent table's name, because that name carries identity, and the fold happens later, when the name is emitted. `AliasState.tableDef` registers the folded form as the alias and collision-checks against the folded form, while keeping its map keyed on the unfolded name — that shared key is what lets a table and a subquery synthesized over it resolve to one and the same alias.

`AliasState.tableRef` has to fold for a different reason. A column reference correlated to a table defined by an *enclosing* query finds no alias registered under its own result path, and falls back to reusing the enclosing query's alias. That fallback is load bearing: registering a fresh alias there instead names a table that is absent from the FROM clause, which breaks nine existing paging tests. Since the enclosing alias is now folded, the fallback has to fold too, or the reference emits `qualified.country` against a FROM entry reading `qualified_country`.

`SubqueryRef.identifier` folds as well, since by then its name is a composed string rather than a `TableName`.

Making the alias slot the authoritative fold point is what turns the rule from a per-call-site convention into something structural: a name reaching an alias can no longer arrive unfolded, whatever composed it.

## Behavioural change worth noting

Because the fold now happens where the alias is emitted rather than where a name is minted, a schema-qualified table renders with an explicit alias wherever it appears, not only under a union:

```sql
FROM qualified.city AS qualified_city
```

That is the Postgres rendering; MSSQL and Oracle take a bare alias with no `AS`, and Oracle rejects `AS` in that position outright.

The consequence for anyone already mapping qualified tables is that the emitted SQL changes globally, not just in the shapes that were broken. Column references that previously read `public.country.code` now read `public_country.code`, because they are prefixed with the alias rather than the qualified name. The results are unchanged; anything asserting on generated SQL text is not.

Mappings whose table names carry no dot are unaffected, and provably so: `foldToIdentifier` replaces `.` with `_`, so for a dot-free name it is the identity, and every site above receives exactly the string it received before.

## Collision safety

Folding a qualifier raises the obvious question: can `qualified.country`, folded to `qualified_country`, collide with an unrelated real table actually named `qualified_country`? `AliasState` deduplicates at render time against every name it has already emitted, regardless of whether that name came from a real table or from a folded synthesized one, and the collision check now runs on the folded form specifically. Whichever of the two is rendered second is uniquified by the shared alias counter.

The fixture contains such a table deliberately, and two tests put both into a single statement — one flat, one with the folded-name table acting as an enclosing table that has a paged child of its own, so it is the side that loses the alias slot and is still referenced from a deeper result path. Both return correct rows.

One pre-existing edge case remains out of scope: a table name containing a literal dot inside a quoted identifier, rather than as a schema separator, was already broken before this change.

## Tests

A shared `SqlQualifiedNamesMapping` and `SqlQualifiedNamesSuite` over a `CREATE SCHEMA qualified` fixture, wired into both `doobie-pg` and `skunk`, eight tests:

- the reported reproducer, a recursive Country → City → Country traversal;
- a top-level limit with a child join, and a nested limit, both exercising `syntheticName`;
- an associative-field query, exercising the `_assoc` alias;
- a nested limit over a union, exercising the identity-bearing subquery name;
- the coexistence pin described above;
- offset/limit paging where a parent table's key column is mapped at the child's result path;
- the same paging shape with the qualified table reached at two different result paths and the folded-name table given a paged child.

Plus a `TableNameSuite` covering `TableName` itself: the split on the last dot, `sqlRef`, `identifier`, and the `toString` contract.

The first six were run against `19453318`, the commit this branch forks from: all six fail there, every one of them with `syntax error at or near "."`. The two paging tests target the `tableRef` and `subqueryToWithQuery` folds specifically; reverting either one alone turns them red, with `invalid reference to FROM-clause entry for table "country"` and a syntax error at the dot in `WITH qualified.country_base` respectively.

Oracle and MSSQL are deliberately not wired into the shared suite; their schema models differ enough that the fixture would not transfer directly, as noted in a comment on the suite.

## Verification

`coreJVM` 557/557, `doobiepg` 261/261, `skunkJVM` 262/262, `doobieoracle` 248/248, `doobiemssql` 248/248, and `headerCheckAll` plus `scalafmtCheckAll` clean. Run locally on 2.13 against freshly recreated containers; CI covers the cross-build.

## Notes for review

MSSQL's `encapsulateUnionBranch` is the one changed site with no direct behavioural test, for the fixture reason above — it is the same pattern as the sql-core sites and was changed by inspection. Flagging it explicitly rather than leaving it to be noticed.

`identifier` is a convention rather than a guarantee: `WithRef.identifier` and `DerivedTableRef.identifier` both return a string verbatim without folding, and are correct only because of what reaches them. A distinct type for alias-position strings, constructible only through the fold, would make the whole class of defect a compile error rather than a test failure. I have not built it here, since it is a wider change than this fix warrants, but this PR fixed two instances of that class and introduced and then fixed a third, so it seems worth recording.

`tlBaseVersion` is bumped to 0.31 because the structured `TableName` changes the sql-core public API, and 0.30.0 is now released, so the check runs against that. No MiMa exclusion filters. #893 currently bumps to 0.30 and will need the same move; whichever of the two lands second needs a trivial rebase.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -36,7 +36,7 @@ ThisBuild / scalaVersion := Scala2
 ThisBuild / crossScalaVersions := Seq(Scala2, Scala3)
 ThisBuild / tlJdkRelease := Some(11)
 
-ThisBuild / tlBaseVersion := "0.30"
+ThisBuild / tlBaseVersion := "0.31"
 ThisBuild / startYear := Some(2019)
 ThisBuild / licenses := Seq(License.Apache2)
 ThisBuild / developers := List(
--- a/modules/doobie-core/src/main/scala/DoobieMapping.scala
+++ b/modules/doobie-core/src/main/scala/DoobieMapping.scala
@@ -58,7 +58,7 @@ trait DoobieMappingLike[F[_]] extends Mapping[F] with SqlMappingLike[F] {
       implicit tableName: TableName,
       typeName: TypeName[T],
       pos: SourcePos): ColumnRef =
-    ColumnRef(tableName.name, colName, (codec, nullable), typeName.value, pos)
+    ColumnRef(tableName, colName, (codec, nullable), typeName.value, pos)
 
   implicit def Fragments: SqlFragment[Fragment] =
     new SqlFragment[Fragment] {
--- a/modules/doobie-mssql/src/main/scala/DoobieMSSqlMapping.scala
+++ b/modules/doobie-mssql/src/main/scala/DoobieMSSqlMapping.scala
@@ -91,7 +91,7 @@ trait DoobieMSSqlMappingLike[F[_]] extends DoobieMappingLike[F] with SqlMappingL
     else
       // The subquery name lands in alias position, so a schema-qualified table name must be
       // folded to a bare identifier first (issue #342).
-      s.toSubquery(TableName.asIdentifier(s.table.name) + "_encaps", Laterality.NotLateral)
+      s.toSubquery(s.table.identifier + "_encaps", Laterality.NotLateral)
 
   def mkLateral(inner: Boolean): Laterality =
     Laterality.Apply(inner)
--- a/modules/doobie-mssql/src/main/scala/DoobieMSSqlMapping.scala
+++ b/modules/doobie-mssql/src/main/scala/DoobieMSSqlMapping.scala
@@ -88,7 +88,10 @@ trait DoobieMSSqlMappingLike[F[_]] extends DoobieMappingLike[F] with SqlMappingL
 
   def encapsulateUnionBranch(s: SqlSelect): SqlSelect =
     if (s.orders.isEmpty) s
-    else s.toSubquery(s.table.name + "_encaps", Laterality.NotLateral)
+    else
+      // The subquery name lands in alias position, so a schema-qualified table name must be
+      // folded to a bare identifier first (issue #342).
+      s.toSubquery(TableName.asIdentifier(s.table.name) + "_encaps", Laterality.NotLateral)
 
   def mkLateral(inner: Boolean): Laterality =
     Laterality.Apply(inner)
--- a/modules/doobie-pg/src/test/scala/DoobiePgSuites.scala
+++ b/modules/doobie-pg/src/test/scala/DoobiePgSuites.scala
@@ -213,6 +213,10 @@ final class QualifiedNamesSuite extends DoobiePgDatabaseSuite with SqlQualifiedN
   lazy val mapping = new DoobiePgTestMapping(transactor) with SqlQualifiedNamesMapping[IO]
 }
 
+final class TableNameSuite extends DoobiePgDatabaseSuite with SqlTableNameSuite {
+  lazy val mapping = new DoobiePgTestMapping(transactor) with SqlQualifiedNamesMapping[IO]
+}
+
 final class RecursiveInterfacesSuite
     extends DoobiePgDatabaseSuite
     with SqlRecursiveInterfacesSuite {
--- a/modules/doobie-pg/src/test/scala/DoobiePgSuites.scala
+++ b/modules/doobie-pg/src/test/scala/DoobiePgSuites.scala
@@ -209,6 +209,10 @@ final class ProjectionSuite extends DoobiePgDatabaseSuite with SqlProjectionSuit
   lazy val mapping = new DoobiePgTestMapping(transactor) with SqlProjectionMapping[IO]
 }
 
+final class QualifiedNamesSuite extends DoobiePgDatabaseSuite with SqlQualifiedNamesSuite {
+  lazy val mapping = new DoobiePgTestMapping(transactor) with SqlQualifiedNamesMapping[IO]
+}
+
 final class RecursiveInterfacesSuite
     extends DoobiePgDatabaseSuite
     with SqlRecursiveInterfacesSuite {
--- a/modules/doobie-pg/src/test/scala/DoobiePgSuites.scala
+++ b/modules/doobie-pg/src/test/scala/DoobiePgSuites.scala
@@ -217,6 +217,10 @@ final class QualifiedNamesSuite extends DoobiePgDatabaseSuite with SqlQualifiedN
   lazy val mapping = new DoobiePgTestMapping(transactor) with SqlQualifiedNamesMapping[IO]
 }
 
+final class TableNameSuite extends DoobiePgDatabaseSuite with SqlTableNameSuite {
+  lazy val mapping = new DoobiePgTestMapping(transactor) with SqlQualifiedNamesMapping[IO]
+}
+
 final class RecursiveInterfacesSuite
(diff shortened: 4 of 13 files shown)


milessabin (changes requested):
I think the "deeper fix" described in the design note would be the right way to go.

phdoerfler:
Fair enough. I'll just close this PR then. I figured it might be worth having this superficial fix in the meantime so the user does not get an unexpected 500 but I do also think a more thorough fix is the way to go ultimately.

phdoerfler:
nvm my last comment. I confused something. Yes, I also think an own `TableName` type would be great and I think it should be a reasonably sized change.

milessabin (changes requested):
Aside from a compilation warning ... LGTM!

phdoerfler:
Thanks for the review and the reproducer. Should be fixed now. However, doing so uncovered a regression in the earlier fold. That should be fixed now as well. I added tests and updated the PR description.

milessabin (approved):
LGTM ... just needs a rebase ... thanks!

PR #873, opened 2026-07-14
Add MariaDB backend

phdoerfler:
This adds MariaDB as a backend, built as a close fork of the MySQL backend (#872). Not only that, it also sits on top of that PR since MariaDB is basically identical to MySQL and, for now, we can just reuse the MySQL test data files.

To review only the MariaDB additions: `git diff topic/mysql-backend topic/mariadb-backend`, and once #872 merges this diff collapses to them by itself.

## No LATERAL

MariaDB has no `LATERAL` derived tables in any released version, unlike MySQL, which has had them natively since 8.0.14. This backend therefore uses the same `NotLateral` machinery SQLite and H2 use rather than MySQL's native rendering — one structural dialect change (`mkLateral`), and the only one. The `SqlMappingLike` hooks it needs are on `main` from #917.

## Driver license

`mariadb-java-client` is LGPL 2.1, a different and more permissive license than the MySQL backend's `mysql-connector-j` (GPLv2 with the Universal FOSS Exception).

## Ascribed nulls for JSON columns

MariaDB's `JSON` type is a `LONGTEXT` alias rather than a native CAST target, so `CAST(NULL AS JSON)` is a syntax error (`ERROR 1064`) where MySQL accepts it. That case falls through to a bare `NULL` here, the treatment `FLOAT`/`DOUBLE` already get, and safe because an ascribed NULL is only ever an inference hint.

Nothing in the shared suites reaches that branch. None of them route a nullable JSON field through the ascription path so the rendering is pinned by a test instead. This PR adds the MariaDB one, asserting a bare `NULL`; its MySQL counterpart, asserting `CAST(NULL AS JSON)`, is in #872. The two dialects are meant to disagree, and a test each is what says so.

## Test data

The `mariadb` docker-compose service mounts `./target/testdata/mysql/`. There is no `mariadb` dialect and no `mariadb.sql` in any dataset. If a real incompatibility ever turns up we can cross that bridge then, either by duplicating the scripts wholesale or by adding a `mariadb.sql` only for the datasets that need one.

MariaDB's client does not have MySQL's latin1-handshake charset problem, so the `entrypoint.sh` workaround the MySQL service needs has no counterpart here.

## Version floor

The service defaults to `mariadb:${MARIADB_VERSION:-lts}`, and the official image's `lts` tag tracks MariaDB's current LTS release (12.3 as of writing), so there is no version number to keep re-pinning. Unlike MySQL, no single feature forces a particular floor here, so the floor is the current LTS this was developed and tested against.

## Disclaimer

This work was made in conjunction with an LLM.


milessabin:
This will need rebasing.

milessabin:
scalafmt and needs to be set "Ready for review".

phdoerfler:
@milessabin This is now ready for review. I recommend merging the MySQL one first and then this one, because the MariaDB PR includes MySQL and doing it in this order will naturally collapse the diffs. Also I don't know what Github does if this is merged first and then the MySQL PR will have 0 new commits to contribute.

milessabin (approved):
LGTM!

PR #876, opened 2026-07-17
Validate schema root operation types

### LINKED ISSUE #174: empty schema should fail at  compile time
I'm just starting to learn about grackle and  a test I discovered that a schema definition cannot be empty. However you only get a failure at runtime. Should it fail at compile time?

```
object ItcMapping extends GenericMapping[Id] {
 val schema = schema"""
 """

 val typeMappings = Nil
}
```

phdoerfler:
Fixes #174.

## Problem

A schema with no `query` root type — either because it has no `type Query` at all and no explicit `schema { }` block, or because an explicit `schema { }` block references a root type that doesn't exist — currently either:

- crashes with an unhelpful `NoSuchElementException` the first time `.queryType` is accessed (e.g. on the first query execution), or
- silently produces a dangling `TypeRef` for an undefined `mutation` / `subscription` root type, since `schema { }` root-type references were never existence-checked the way ordinary field/type references are.

Per the [GraphQL spec, "Root Operation Types"](https://spec.graphql.org/October2021/#sec-Root-Operation-Types):

> The `query` root operation type must be provided and must be an Object type. The `mutation` root operation type is optional [...] If it is provided, it must be an Object type. Similarly, the `subscription` root operation type is also optional [...] If it is provided, it must be an Object type.

This should be a schema validation error, not a runtime crash or silent bad state.

## Fix

Adds `SchemaValidator.validateRootTypes`, wired into the existing `validateSchema` chain right after `validateReferences`. For each of `query` (mandatory), `mutation` (optional), `subscription` (optional):

- missing and mandatory → `No 'query' root operation type in schema`
- present but undefined → `Undefined type '<name>' specified as '<op>' root operation type`
- present, defined, but not an Object type → `Type '<name>' specified as '<op>' root operation type is not an object type`

Because both the `schema"""..."""` compile-time macro and the runtime `Schema(string)` constructor share the same `SchemaParser → validateSchema` pipeline, this closes the gap at construction time for both entry points. No changes needed to `queryType`/`mutationType`/`subscriptionType` themselves.

## Test plan

- Three new targeted tests in `SchemaSuite.scala`, one per error path (missing query, dangling reference, non-object root type).
- ~12 pre-existing test fixtures that relied on Query being optional for schema *construction* (to unit-test unrelated validators in isolation) updated to include a minimal `type Query { foo: Int }`.

### DIFF
--- a/modules/core/src/main/scala/schema.scala
+++ b/modules/core/src/main/scala/schema.scala
@@ -1970,6 +1970,7 @@ object SchemaValidator {
       defns: List[TypeDefinition],
       typeExtnDefns: List[Ast.TypeExtension]): List[Problem] =
     validateReferences(schema, defns) ++
+      validateRootTypes(schema) ++
       validateUniqueDefns(schema) ++
       validateUniqueFields(schema) ++
       validateUnionMembers(schema) ++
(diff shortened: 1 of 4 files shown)


phdoerfler:
Rebased onto current main. This will fail the build until #899 lands: `v0.30.0` is tagged but main's `tlBaseVersion` is still `0.29`, so sbt-typelevel rejects any branch with a commit on top of the tag. Verified locally that with that bump in place this branch is green, MiMa included.

hugo-vrijswijk (approved):
Very nice. Thanks! There are some cases where an invalid schema could still be made (`type Query { foo: Int } extend schema { query: Undefined }` or `extend schema { query: Bar }; scalar Bar`), but that seems very unlikely to actually occur. So I'm OK merging like this

phdoerfler:
> Very nice. Thanks! There are some cases where an invalid schema could still be made (`type Query { foo: Int } extend schema { query: Undefined }` or `extend schema { query: Bar }; scalar Bar`), but that seems very unlikely to actually occur. So I'm OK merging like this

Sounds like a good enhancement to be covered by a future PR 👍

phdoerfler:
@hugo-vrijswijk I took the liberty and filed https://github.com/typelevel/grackle/issues/903 so it's not lost

PR #879, opened 2026-07-25
Make QuerySizeValidator's combined-limit case reachable; add docs

rpiotrow:
The depth && width case was ordered after the individual limit cases, so "Query is too complex" could never be reported. Reorder the cases, cover all three failure branches with tests, add Scaladoc, and add a "Compiler Phases" tutorial page showing how to enable the validator via compilerPhases.


milessabin (changes requested):
LGTM, but could you fix the header/fmt issues. Thanks!

milessabin (approved):
LGTM!

PR #885, opened 2026-08-02
fix: avoid BigDecimal.equals exception cost in FailedJoin/Table checks

### LINKED ISSUE #884: Expensive calls to BigDecimal.equals
Grackle seems to call `BigDecimal.equals` a lot. In my case, these take 50% of the thread's time which processes the query (490 of 1091 samples) (see highlighted region):
Image

One of, if not the main culprit, seems to be `assert(leafFocus != FailedJoin)` in SqlMapping.
Replacing the equals with eq and thus referential equality (which should be safe here since FailedJoin is a singleton) there and in a few other related areas gets rid of the costly BigDecimals.equals completely, avoiding the expensive stack trace creation.

phdoerfler:
Fixes #884.

Fun fact: This is a `scala.math.BigDecimal`-specific problem, not a JVM one. `java.math.BigDecimal.equals` doesn't do this, only Scala's wrapper does.

## Fix

Swapped the `FailedJoin`/`None` sentinel checks in `Table.select`/`definesAll`/`filterDefined`/`group`/`count` and the `SqlCursor.field` assert for reference-equality checks instead (`FailedJoin.isFailedJoin`, `Table.isNone`, both just `v.asInstanceOf[AnyRef] eq <sentinel>`). Should be exact since both `FailedJoin` and `None` are singletons, and it means `BigDecimal`'s overridden `equals` never gets called for these checks. Went through each changed line by hand to make sure nothing's behaving differently.

An alternative fix would be to use a BigDecimal-specific type check (`case _: BigDecimal => eq check; case _ => ==`). One might argue that, after all, this is just a workaround for a specific Scala limitation in `BigDecimal`.
Or one might argue that in the end it all boils down to reference equality anyway since we are comparing sentinel values. Treating it as such also makes the check more robust. The next weird `equals` of a mapped column will no longer impact performance in this way. Also, there is a minor performance aspect to this: Making this into a BigDecimal-workaround would introduce additional type checks at runtime (and while fast, they would run often).
So, I went with the universal reference equality + a comment explaining where it all started.

I also didn't limit the fix to `SqlCursor.field`'s assert, even though that's the one that showed up in the profiler. `Table.select`/`definesAll`/`filterDefined`/`group`/`count` all had the exact same pattern, all on the same row-assembly hot path, and this is shared core code used by every SQL backend (Postgres, MySQL, MSSQL, H2, ...). Scoping the fix to just the one line the profiler happened to flag would've left the rest of the bug in place.

I used async-profiler to discover the issue and to verify it was gone after applying the fix. As the test data set, I used Microsoft's AdventureWorks data set. This was all found as part of a benchmark I am writing, and the data set was chosen specifically for that benchmark. To avoid relying solely on the profiler output, I also added instrumentation (I might file that as a separate PR if that seems useful), which reliably showed that the "un-flattening", or row-assembly, phase dropped by about 50%, from ~900ms down to ~450ms.

### DIFF
--- a/modules/sql-core/src/main/scala/FailedJoin.scala
+++ b/modules/sql-core/src/main/scala/FailedJoin.scala
@@ -16,6 +16,18 @@
 package grackle.sql
 
 /**
- * A sentinal value representing the empty column values from a failed join.
+ * A sentinel value representing the empty column values from a failed join.
  */
-case object FailedJoin
+case object FailedJoin {
+
+  /**
+   * Cheap equality check for the `FailedJoin` sentinel.
+   *
+   * Column values are `Any` and may be a `scala.math.BigDecimal`, whose `equals` throws and
+   * catches an `ArithmeticException` whenever compared to a value of another type (see
+   * `BigDecimal.isValidLong`). `a == b` calls `a.equals(b)`, so keeping `FailedJoin` on the
+   * left runs the singleton's own cheap `equals` instead of the column value's -- do not swap
+   * the operands here.
+   */
+  def isFailedJoin(v: Any): Boolean = FailedJoin == v
+}
--- a/modules/sql-core/src/main/scala/SqlMapping.scala
+++ b/modules/sql-core/src/main/scala/SqlMapping.scala
@@ -31,6 +31,7 @@ import grackle.Predicate._
 import grackle.Query._
 import grackle.ValidationFailure.Severity
 import grackle.circe.CirceMappingLike
+import grackle.sql.FailedJoin.isFailedJoin
 import grackle.syntax._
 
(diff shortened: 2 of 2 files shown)


phdoerfler:
I have pushed an update with a simpler fix. All that is actually needed is to swap the sides of `.equals` to avoid the costly `BigDecimal.equals` — `isFailedJoin`/`isNone` now just keep the sentinel on the left of `==` instead of using `asInstanceOf`/`eq`. Same effect, smaller diff, call sites unchanged.

The bare `case FailedJoin =>` / `case None =>` arms are back to their original form too — never needed the `eq`-guard.

milessabin (approved):
LGTM!

PR #886, opened 2026-08-03
Don't report a shared fragment as a fragment cycle

toddburnside:
`validateVariablesAndFragments` accumulated visited fragments into a single `seen` set spanning the whole traversal rather than tracking the fragments on the current path. A fragment reachable by more than one path was therefore reported as a cycle, e.g.

    fragment a on T { x { ...c } y { ...b } }
    fragment b on U { z { ...c } }
    fragment c on V { w }

failed with "Fragment cycle starting from 'a'".

Replace the check with a proper DFS: `path` holds the fragments on the current traversal path and `done` those already explored in full, a fragment joins `done` only after its references have been explored, and the `path` check precedes the `done` check. Genuine recursion is still detected, and reported against the same fragment as before.

Claude-Session: https://claude.ai/code/session_01Lyt6Er1f7k4DDcNBoZEUGS

### DIFF
--- a/modules/core/src/main/scala/compiler.scala
+++ b/modules/core/src/main/scala/compiler.scala
@@ -445,34 +445,36 @@ class QueryCompiler(parser: QueryParser, schema: Schema, phases: List[Phase]) {
       val fragRefs: Map[String, (Set[String], Set[String])] =
         frags.map { frag => (frag.name, collectQueryRefs(frag.child)) }.toMap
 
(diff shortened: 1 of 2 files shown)


tpolecat (approved):
This looks good to me. :+1: 

It's a blocking issue for us so it would be good to have this change.

milessabin (approved):
LGTM!

PR #890, opened 2026-08-07
Render descriptions in `Schema.toString`

hugo-vrijswijk:
Descriptions were previously ignored for many types when rendering a `Schema` to a string. This commit adds support for rendering descriptions, using either quoted strings or block strings as appropriate.


tpolecat (approved):

milessabin (approved):
LGTM!

PR #891, opened 2026-08-08
Raise munit timeout for database-backed test suites

phdoerfler:
munit-cats-effect times a test out after 30 seconds, which isn't much headroom for the database-backed suites on a constrained machine. sbt runs each backend's tests in its own forked JVM and several of those contend for a single docker daemon, so a test can exceed 30s for reasons that have nothing to do with what it asserts. It surfaces as a `TimeoutException` in whichever suite happens to run first after the containers come up. I've seen it on `Embedding2Suite.paging`, `Paging3Suite.paging (initial)` and skunk's `TreeSuite.root query`, each of which passes on a rerun of the identical commit.

This raises the timeout to two minutes for the backend suites. It's a single shared `SqlDatabaseSuite` base in sql-core's test scope rather than an override on each of the per-backend traits, because `DoobiePgDatabaseSuite` mixes in both `DoobieDatabaseSuite` and `SqlPgDatabaseSuite` and two separate overrides of `munitIOTimeout` would conflict.

The commit has been riding along in my four open backend PRs (#864, #866, #872, #873), which each carry a copy of it. It stands on its own and is useful to anyone running the suites locally, so it seemed better raised separately; those PRs can drop their copy at their next rebase.

### DIFF
--- a/modules/doobie-core/src/test/scala/DoobieDatabaseSuite.scala
+++ b/modules/doobie-core/src/test/scala/DoobieDatabaseSuite.scala
@@ -18,13 +18,12 @@ package grackle.doobie.test
 import java.time.Duration
 
 import io.circe.{Decoder => CDecoder, Encoder => CEncoder}
-import munit.CatsEffectSuite
 import org.typelevel.doobie.Meta
 
 import grackle.doobie.DoobieMappingLike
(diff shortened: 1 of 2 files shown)


milessabin (approved):
LGTM!

PR #893, opened 2026-08-08
Don't drop rows for a non-null field under a nullable parent

### LINKED ISSUE #888: Non-null field under a nullable parent produces an INNER JOIN that drops rows
It appears that there is a bug in Grackle's SQL generation. If a GraphQL query goes through a nullable JOIN field and queries something past it, it can lead to objects being dropped from the result thanks to an INNER JOIN that eliminates rows produced by a preceding LEFT JOIN. The join-type decision lives in the shared `sql-core`, so this should affect every SQL backend, though I have only reproduced it on Postgres. I tasked the LLM with creating a small reproducer branch and a detailed issue description (with some edits here and there):

---

A non-null field nested beneath a nullable parent generates an `INNER JOIN` in the same flat join chain as the parent's `LEFT JOIN`, so rows whose nullable parent is absent are eliminated. They should be returned with the parent as `null` — the child's non-null-ness only applies when the parent exists.

Failing test: [`modules/doobie-pg/src/test/scala/NullableParentInnerJoinSuite.scala`](https://github.com/phdoerfler/grackle/blob/repro/nullable-parent-inner-join/modules/doobie-pg/src/test/scala/NullableParentInnerJoinSuite.scala) on [`phdoerfler/grackle@repro/nullable-parent-inner-join`](https://github.com/phdoerfler/grackle/tree/repro/nullable-parent-inner-join), branched off current `main`. It creates and drops its own tables, so it needs only a running `postgres` service and no changes to `testdata/pg`:

```
docker compose up -d postgres
sbt "doobiepg/testOnly *NullableParentInnerJoinSuite"
```

That branch also carries a small http4s server (`sbt "demo/runMain demo.nullableparent.NullableParentServer"`) serving the same schema on `/nullrepro`, if it is easier to poke at by hand.

Schema:

```graphql
type Query { as: [A!]! }
type A { name: String!, b: B }     # nullable
type B { name: String!, c: C! }    # non-null
type C { name: String! }
```

Data: two `A` rows, one with a `b`, one with `b_id` NULL.

`{ as { name b { name c { name } } } }` returns only `a-with-b`. Expected `a-without-b` as well, with `"b": null`.

Per [§6.4.3 Value Completion](https://spec.graphql.org/October2021/#sec-Value-Completion), step 2 of `CompleteValue()` reads:

> If `result` is `null` (or another internal value similar to `null` such as `undefined`), return `null`.

Completing `A.b` for the row with a NULL `b_id` stops there: `B` is not a Non-Null type so step 1 does not apply, and step 2 returns `null`. Only step 5 executes sub-selections, and it is never reached — so `{ name c { name } }` is never executed and `c: C!` is never resolved. Its non-null-ness cannot constrain anything, because nothing asks for it.

This is distinct from [Errors and Non-Null Types](https://spec.graphql.org/October2021/#sec-Errors-and-Non-Null-Fields), which governs a non-null field that genuinely *does* resolve to null and propagates an error to the nearest nullable ancestor. That rule does not apply here, since `c` is never resolved.

Worth noting where in execution this goes wrong: `as: [A!]!` completes via step 3, "a list where each list item is the result of calling `CompleteValue(innerType, fields, resultItem, …)`, where `resultItem` is each item in `result`" — one completed item per resolved item. So the missing row is not a value-completion decision at all; field resolution returned fewer items than it should have. The defect is in the emitted SQL, upstream of any null-handling semantics.

Emitted SQL:

```sql
FROM nulltest_a
LEFT JOIN nulltest_b ON ((nulltest_b.id = nulltest_a.b_id))
INNER JOIN nulltest_c ON ((nulltest_c.id = nulltest_b.c_id))
```

Each join is correct alone — `b` nullable so `LEFT`, `c` non-null so `INNER` — but they share one flat chain, so the row with a NULL `b_id` gets NULL `b` columns from the left join and is then eliminated by the inner join beneath it.

**The suite includes a control** that isolates it: the same mapping and data, queried as `{ as { name b { name } } }` without descending into `c`, returns both rows. So the mapping is sound and the trigger is specifically a non-null object nested under a nullable one.

The join type comes from `SqlSelect.nest`:

```scala
// modules/sql-core/src/main/scala/SqlMapping.scala:2378
val inner = !context.tpe.isNullable && !context.tpe.isList
```

That boolean is passed to `Join.toSqlJoin` (lines 2443, 2451, 2459, 2483), stored on `SqlJoin`, and rendered as the keyword at line 3486: `val kind = if (inner) "INNER" else "LEFT"`. It is derived from the field's own type; `nest` receives a `parentContext` but never consults it, so a nullable ancestor along the join path is not accounted for.

Two local experiments confirm that line is what drives the behaviour, rather than merely sitting on the path. Forcing `val inner = false` makes the repro pass. Adding `&& !parentContext.tpe.isNullable` also makes it pass, and leaves the rest of the suite unchanged: `doobiepg/test` goes from 1 failure to 0, the difference being the repro itself.

This is first and foremost a clue, not as a proposed patch. Passing the existing suite only shows the suite does not cover whatever such a change might break — and it demonstrably does not cover this shape either, which is how the bug survived. Whether the right remedy is widening that condition or restructuring the nesting into a subquery is your call.

No existing suite covers the shape: `SqlTreeMapping` is the only nullable to-one, and beneath it there is only a scalar, never a non-null object.

Tested against `main` at tlBaseVersion 0.29, Postgres via doobie.

### LINKED ISSUE #892: On MSSQL an `OUTER APPLY` is not correlated, so a nullable lateral join behaves as an inner join
## On how this was written

This issue report was mostly done by an LLM. I have poked and prodded at it, encouraged it to verify rather than assume, enforced a proper software engineering workflow including TDD and reviews on top of which I did a manual review. It was an iterative process. The issue is now at a stage where I can't find anything obviously wrong with it but I will admit that the SQL generation is a bit arcane and things might have slipped past me.

## The Issue

MSSQL has no `LATERAL` keyword, so `DoobieMSSqlMapping.mkLateral` renders a lateral join as `CROSS`/`OUTER APPLY`. `APPLY` takes no `ON` clause, so the join's conditions can't be rendered with the join and are instead lifted into the enclosing select's `WHERE` clause by `Laterality.Apply.joinPredicates`, via `mkNested`'s `extraWheres`.

For `CROSS APPLY` that's equivalent to an `ON` clause. For `OUTER APPLY` it isn't: the `WHERE` clause is applied after the join, and discards precisely the null padded rows the outer join produced. The subquery is never correlated to the enclosing select at all — it contains no reference to the outer table — so the `APPLY` behaves as a cross join whose result is then filtered down to matching rows. An `OUTER APPLY` is therefore an inner join in all but name.

## Reproducing it

Any query which puts a nullable field over a lateral subquery will do. The one which surfaced this uses the schema from #888 — `type A { name: String!, b: B }` with `b` nullable, and `type B { name: String!, c: C! }` with `c` non-null — over three rows of `A`, one with a matching `B`, one whose `b_id` dangles, and one whose `b_id` is null:

```graphql
query { as { name b { name c { name } } } }
```

MSSQL emits:

```sql
SELECT nested.id, nested.id_alias_0, nested.name, nested.name_alias_1,
       nullable_parent_a.id AS id_alias_2, nullable_parent_a.name AS name_alias_3
FROM nullable_parent_a
OUTER APPLY (
  SELECT nullable_parent_b.id AS id_alias_0, nullable_parent_c.id, nullable_parent_c.name,
         nullable_parent_b.name AS name_alias_1
  FROM nullable_parent_b
  INNER JOIN nullable_parent_c ON (nullable_parent_c.id = nullable_parent_b.c_id)
) nested
WHERE (nullable_parent_a.b_id = nested.id_alias_0)
```

which returns one row. Postgres emits the same subquery as a `LEFT JOIN LATERAL ... ON (...)` and returns all three, two of them with `"b": null`. Executing the MSSQL statement directly confirms it isn't a mapping or decoding artefact: as written it yields one row, and moving the correlation inside the `APPLY` and dropping the outer `WHERE` yields three.

## Note

This is independent of #888 and older than it. #888 needs a specific schema shape to bite; this affects any lateral join under a nullable field on MSSQL. It only became visible through #888 because before that fix the nested select was flattened into the join chain and no `APPLY` was emitted at all, so the two have to be fixed together for either to be observable.

There's a neighbouring case this doesn't cover, left alone deliberately: the inner `mkNested` runs with `outer = !multiTable`, so in the `multiTable` path an `APPLY` join built there contributes no join conditions at all — the guard sits ahead of the distribution either way, so this is unchanged by a fix. Nothing in the suite fails on it, though that's the absence of a witness rather than evidence there isn't one.

phdoerfler:
This PR was mostly done by an LLM. I have poked and prodded at it, encouraged it to verify rather than assume, enforced a proper software engineering workflow including TDD and reviews on top of which I did a manual review. It was an iterative process. The PR is now at a stage where I can't find anything obviously wrong with it but I will admit that the SQL generation is a bit arcane and things might have slipped past me.

And now, without further ado, the detailed LLM analysis:

---

Fixes #888. Also fixes #892, which #888 turned out to depend on.

Three defects, one observable symptom. The first is the one #888 describes: a nested select carrying an `INNER` join was flattened into the enclosing select's join chain even when it attached via a `LEFT` join, so the inner join sat below the outer one and eliminated the rows the outer join had null padded. `SqlSelect.nest` now declines to flatten in that case and builds a subquery instead, which keeps the inner join scoped where it's genuinely valid rather than discarding the optimisation.

Fixing that alone had no effect on MSSQL, which turned out to have a second and older defect, described in #892: an `OUTER APPLY` was never correlated to its enclosing select, so it behaved as an inner join and discarded the same rows again for an unrelated reason. `Laterality.joinPredicates` is replaced by `distributeJoinConditions`, which can rewrite the join as well as yield predicates for the enclosing select, and `Apply` uses that to wrap the subquery of an `OUTER APPLY` in a correlating select. Applying the conditions to the subquery's result rather than pushing them into it preserves any `LIMIT`, `OFFSET` or `DISTINCT`, and keeps a same named table inside the subquery from shadowing one of the enclosing select — both of which regressed when tried the other way round.

Correlating once isn't enough, which is the third fix. Nesting a select re-examines joins an earlier nesting already correlated, and `correlate` didn't recognise the wrapper it had itself built — so it declined, the conditions were lifted a second time, and the correlation was undone. It now recognises that wrapper and leaves the join alone. With that, no query in any of the four backends' suites reaches a decline at all.

None of the three is observable without the others, which is why they're here together: without the first no `APPLY` is emitted for these queries at all, without the second the first is a no-op on MSSQL, and without the third the second only holds to a depth of one.

Replacing `joinPredicates` isn't binary compatible, so the last commit bumps `tlBaseVersion` to 0.30. It's kept separate in case you'd rather own that yourself or take it in a different release.

Worth noting that the issue understates the reach. The join is chosen by `!context.tpe.isNullable && !context.tpe.isList`, so a list is attached the same way a nullable field is, and a parent with no children at all was being dropped for the same reason: `[B!]!` with a non-null field beneath `B` lost every parent whose list was empty. That needs no nullable field anywhere in the schema, so it's the easier of the two to hit by accident. There's a test for each shape.

The tests are a shared mapping and suite run on all four SQL backends, with fixture data including a row whose non-null reference dangles — the case that distinguishes a row being absent from a row being dropped — and a parent with an empty list. A third test descends two lists through a non-null field, which is what reaches the repeated correlation; it needs a second reference back to the parent type, so that the inner list can be empty where the outer one isn't.

One thing deliberately out of scope. A genuinely dangling non-null reference is still not reported as an error: before this change the row vanished from the result entirely, and after it the parent row is returned with the offending field as `null`. Both are wrong under §6.4.3 step 1, which asks for an execution error propagated to the nearest nullable ancestor. That's a separate defect with a separate remedy — the object field path at `SqlCursor` narrows without checking, where the interface and union path checks first via `narrowsTo` — and I'd rather raise it on its own than widen this. The tests assert the post-fix behaviour as it stands, so they'll need updating when it's addressed.

Verified on all four backends: `doobiepg`, `doobieoracle` and `skunkJVM` are green, and `doobiemssql` is green apart from three pre-existing `MovieSuite` failures which fail identically at the base commit. Those are a local artifact rather than anything in this change — the container runs in the host's timezone, so `moviesShownBetween` resolves its range boundary an hour out and returns a fourth film — and they're what #862 pins to UTC.

### DIFF
--- a/modules/sql-core/src/main/scala/SqlMapping.scala
+++ b/modules/sql-core/src/main/scala/SqlMapping.scala
@@ -1442,7 +1442,19 @@ trait SqlMappingLike[F[_]] extends CirceMappingLike[F] with SqlModule[F] { self 
     trait Laterality {
       def toFragment: Fragment
       def joinToFragment(join: SqlJoin, subquery: SubqueryRef): Aliased[Fragment]
-      def joinPredicates(join: SqlJoin): List[Predicate]
+
+      /**
+       * Distribute the conditions of `join` between the join itself and the enclosing select.
+       *
+       * Yields a possibly rewritten join, together with the predicates which must be added to
+       * the enclosing select's WHERE clause for the join to have its intended semantics.
+       *
+       * Lateralities which render the join with an `ON` clause need neither, and so yield the
+       * join unchanged and no predicates.
+       */
+      def distributeJoinConditions(
+          join: SqlJoin,
+          subquery: SubqueryRef): (SqlJoin, List[Predicate])
     }
     object Laterality {
       def apply(lateral: Boolean, inner: Boolean): Laterality =
@@ -1452,21 +1464,94 @@ trait SqlMappingLike[F[_]] extends CirceMappingLike[F] with SqlModule[F] { self 
         def toFragment: Fragment = Fragments.empty
         def joinToFragment(join: SqlJoin, subquery: SubqueryRef): Aliased[Fragment] =
           join.toFragmentWithoutLaterality
-        def joinPredicates(join: SqlJoin): List[Predicate] = Nil
+        def distributeJoinConditions(
+            join: SqlJoin,
+            subquery: SubqueryRef): (SqlJoin, List[Predicate]) = (join, Nil)
       }
       case object Lateral extends Laterality {
         def toFragment: Fragment = Fragments.const("LATERAL ")
         def joinToFragment(join: SqlJoin, subquery: SubqueryRef): Aliased[Fragment] =
           join.toFragmentWithoutLaterality
-        def joinPredicates(join: SqlJoin): List[Predicate] = Nil
+        def distributeJoinConditions(
+            join: SqlJoin,
+            subquery: SubqueryRef): (SqlJoin, List[Predicate]) = (join, Nil)
       }
       case class Apply(inner: Boolean) extends Laterality {
         def toFragment: Fragment =
           Fragments.const(if (inner) "CROSS " else "OUTER ") |+| Fragments.const("APPLY ")
         def joinToFragment(join: SqlJoin, subquery: SubqueryRef): Aliased[Fragment] =
           subquery.toDefFragment
-        def joinPredicates(join: SqlJoin): List[Predicate] =
+
+        /**
+         * An `APPLY` join is rendered without an `ON` clause, so its conditions have to be
+         * expressed elsewhere.
+         *
+         * For `CROSS APPLY` they are lifted into the enclosing select's WHERE clause, which for
+         * an inner join is equivalent to an `ON` clause.
+         *
+         * For `OUTER APPLY` that would be wrong: the WHERE clause is applied after the join and
+         * would discard precisely the null padded rows the outer join produced. Instead the
+         * subquery is wrapped in a correlating select which applies the conditions to its
+         * result, referring to the enclosing select's tables. That is legal because `APPLY` is
+         * lateral, and is what makes `OUTER APPLY` equivalent to `LEFT JOIN LATERAL`.
+         *
+         * Only a subquery of the right shape can be correlated, and the conditions are lifted
+         * as a last resort otherwise. For an `OUTER APPLY` that reinstates the semantics
+         * described above, so it has to stay unreachable: the shapes `correlate` declines are
+         * those `addFilterOrderByOffsetLimit` builds, and those join on a predicate, which is
+         * short circuited before it.
+         */
+        def distributeJoinConditions(
+            join: SqlJoin,
+            subquery: SubqueryRef): (SqlJoin, List[Predicate]) =
+          if (inner || join.isPredicate) (join, liftedPredicates(join))
+          else
+            correlate(join, subquery).fold((join, liftedPredicates(join)))(join0 =>
+              (join0, Nil))
+
+        /**
+         * The join's conditions expressed as predicates of the enclosing select.
+         */
+        private def liftedPredicates(join: SqlJoin): List[Predicate] =
           if (!join.isPredicate) join.on.map { case (p, c) => Eql(p.toTerm, c.toTerm) } else Nil
+
+        /**
+         * Yields a copy of `join` with its conditions moved inside its subquery.
+         *
+         * The subquery is wrapped in a select which applies the conditions to the subquery's
+         * result, so that any LIMIT, OFFSET, DISTINCT or window function in the subquery is
+         * evaluated first, exactly as it would be were the conditions in the ON clause of a
+         * `LEFT JOIN LATERAL`. Wrapping also ensures that only the subquery's exposed columns
+         * are in scope where the conditions are applied, so that a table of the enclosing
+         * select can't be shadowed by a same named table within the subquery.
+         *
+         * The subquery side of each condition names a column exposed by the subquery, which has
+         * to be mapped to the corresponding column of the wrapper. Yields `None` if that isn't
+         * possible, in which case the caller falls back to lifting the conditions into the
+         * enclosing select.
+         */
+        private def correlate(join: SqlJoin, subquery: SubqueryRef): Option[SqlJoin] =
+          subquery.subquery match {
+            case sel: SqlSelect if sel.withs.isEmpty =>
+              sel.table match {
+                case table: TableRef =>
+                  val exposed =
+                    join.on.traverse {
+                      case (p, c) => sel.cols.find(_ == c.subst(subquery, table)).map((p, _))
+                    }
+                  exposed.map { exposed0 =>
+                    val wrapper =
+                      sel.toSubquery(subquery.name + "_corr", NotLateral, correlated = true)
+                    val wheres =
+                      exposed0.map {
+                        case (p, c) => Eql(p.toTerm, c.derive(wrapper.table).toTerm)
+                      }
+                    join.copy(child = subquery.copy(subquery = wrapper.copy(wheres = wheres)))
+                  }
+                case _ => None
+              }
+            case _ => None
+          }
       }
     }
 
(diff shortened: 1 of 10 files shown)


modules/sql-core/src/main/scala/SqlMapping.scala
```diff
@@ -1396,21 +1408,92 @@ trait SqlMappingLike[F[_]] extends CirceMappingLike[F] with SqlModule[F] { self
...
+              .getOrElse(throw new SqlMappingException(
+                s"OUTER APPLY subquery '${subquery.name}' could not be correlated"))
+
+        /**
+         * The join's conditions expressed as predicates of the enclosing select.
+         */
+        private def liftedPredicates(join: SqlJoin): List[Predicate] =
           if (!join.isPredicate) join.on.map { case (p, c) => Eql(p.toTerm, c.toTerm) } else Nil
+
+        /**
```
hugo-vrijswijk, line 1512: This throw is not wrapped in any `Result.catchNonFatal` or other exception boundary. Meaning this could throw exceptions not caught by Grackle at all. Can you change either the call, or the signature to catch the exception somewhere?

build.sbt
```diff
@@ -36,7 +36,7 @@ ThisBuild / scalaVersion := Scala2
 ThisBuild / crossScalaVersions := Seq(Scala2, Scala3)
 ThisBuild / tlJdkRelease := Some(11)
 
-ThisBuild / tlBaseVersion := "0.29"
+ThisBuild / tlBaseVersion := "0.31"
```
hugo-vrijswijk, line 39: I think bumping this can be prevented. The `correlate` check is only used in one place where it could be replaced with `wrapped.name == correlationName(subquery)`. So if the check is done there instead the `correlate` boolean can be removed and signatures kept the same. The `distributeJoinConditions` could be put in the Laterality companion object (just like the `correlate`) and with a few moves there shouldn't be any bincompat changes

modules/sql-core/src/main/scala/SqlMapping.scala
phdoerfler: Good catch! The call to `mkJoins` is now wrapped in `Result.catchNonFatal`.

build.sbt
phdoerfler: True, but a) I don't think there's much reason to avoid the bump. Grackle is pre-1.0, and b) #871 already bumped `tlBaseVersion` to 0.31 in the meantime, so this is moot now. The companion-object move would work, but I think that would be at a slight expense of readability.

I'd also rather keep the `correlated` field on `SubqueryRef` than compare against `correlationName(subquery)`. The field states the invariant directly, where the name comparison infers it from a synthesized string, and if that inference ever stops holding it fails silently: `correlate` returns the join unchanged and the `OUTER APPLY` is left uncorrelated again.

hugo-vrijswijk (approved):

PR #900, opened 2026-08-19
Accept names that start with a keyword

hugo-vrijswijk:
The fragment-name parser rejected any name that starts with "on". The enum-value parser rejected any name that starts with "true", "false" or "null". Both lookaheads had no word-boundary check.

Changes:
- `FragmentName` and `EnumValue` now reject only the exact keyword.
- `BooleanValue` requires a word boundary after "true" or "false", and backtracks so that `trueStory` can parse as an enum value.
- `NullValue` backtracks for the same reason.
- `InlineFragment` backtracks its type condition, so that `...onFoo` falls through to the fragment-spread branch.


milessabin (approved):
LGTM!

PR #901, opened 2026-08-19
Accept a comment at the end of a file

hugo-vrijswijk:
The comment parser required a line terminator after the comment text. A comment at the end of the file has none, so the whole document failed to parse.

The terminator is now `charIn('\n', '\r') | Parser.end`.

### DIFF
--- a/modules/core/src/test/scala/parser/ParserSuite.scala
+++ b/modules/core/src/test/scala/parser/ParserSuite.scala
@@ -971,6 +971,15 @@ final class ParserSuite extends CatsEffectSuite {
     assertEquals(resFail, Result.failure(expectedFail))
   }
 
+  test("comment at end of file without a line terminator") {
+    val expected =
+      List(Operation(Query, None, Nil, Nil, List(Field(None, Name("x"), Nil, Nil, Nil))))
+
+    assertEquals(parser.parseText("query { x } # done"), Result(expected))
+    assertEquals(parser.parseText("query { x } #"), Result(expected))
+    assertEquals(parser.parseText("query { # inner\n x } # a\n# b"), Result(expected))
+  }
+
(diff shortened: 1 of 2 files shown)


modules/core/src/test/scala/parser/ParserSuite.scala
```diff
@@ -971,6 +971,19 @@ final class ParserSuite extends CatsEffectSuite {
...
+  test("comment at end of file without a line terminator") {
+    val expected =
+      List(Operation(Query, None, Nil, Nil, List(Field(None, Name("x"), Nil, Nil, Nil))))
+
+    assertEquals(parser.parseText("query { x } # done"), Result(expected))
+    assertEquals(parser.parseText("query { x } #"), Result(expected))
+    assertEquals(parser.parseText("query { # inner\n x } # a\n# b"), Result(expected))
+  }
+
+  test("comment only document at end of file") {
```
phdoerfler: This asserts that a document with no definitions parses. That behaviour predates this PR, it comes from `Definition.rep0` at line 76, but the spec has `Document ::= Definition+`, so a document with nothing in it should be a parse error rather than an empty document. We might not want to establish something in the tests that violates the spec.

phdoerfler (changes requested):
Aside from the comment-only test it looks good. The GraphQL grammar ends a comment at the end of the input as well as at a `LineTerminator`, so this matches the spec.

modules/core/src/test/scala/parser/ParserSuite.scala
hugo-vrijswijk: Agreed, I'll delete the test. Thanks for catching it

phdoerfler (approved):

PR #902, opened 2026-08-19
Allow `@skip` and `@include` on the same selection

hugo-vrijswijk:
The specification allows both `@skip` and `@include` on one selection. `isSkipped` matched on the full directive list and failed with "skip/include directives must be unique" for the combination.

https://spec.graphql.org/September2025/#sec-Directives-Are-Unique-per-Location


milessabin (approved):
LGTM!

PR #904, opened 2026-08-20
Add union-branch ordering coverage on a schema-qualified table

phdoerfler:
Adds a shared `SqlUnionOrderSuite` that tests a union with `order` and `limit`, run on Postgres, skunk, MSSQL and Oracle against a schema-qualified table (`qualified.union_order_entities`).

#871 changed how qualified table names are folded in MSSQL union-branch encapsulation, but its tests (`SqlQualifiedNamesSuite`) only run on Postgres and skunk, so those MSSQL paths have no test today. The existing `SqlUnionSuite` does not cover order or limit on any backend.

The rows are inserted out of alphabetical order on purpose (id 1 Charlie, 2 Alpha, 3 Bravo, 4 Delta), so a query that returns the first two in insert order cannot pass the check by chance.

A schema is a user in Oracle, so the fixture there needs a second user. `01_create_user.sql` creates it and grants TEST the rights to populate it, including `CREATE ANY INDEX` for the primary key index.

Test only, no library changes.


milessabin (approved):
LGTM!

PR #908, opened 2026-08-21
Fix list coercion, variable usage validation and argument defaults

hugo-vrijswijk:
A non-list value is a valid input for a list type. Only literal lists were accepted, so `arg: [String]` rejected `"foo"` where the specification accepts it as `["foo"]`. A single value is now wrapped, recursively for nested lists, and `null` is left unwrapped. Default values pass through the same coercion, so a supplied value and an equal default no longer produce different results.

A variable with no supplied value must fall back to the default of the location it fills. The default was applied to an absent argument only, not to an absent variable.

A variable was also never compared with the type of the location where it is used. Value coercion rejected some incompatible usages as a side effect, and single value coercion removes that side effect. [Rule 5.8.5](https://spec.graphql.org/September2025/#sec-All-Variable-Usages-Are-Allowed) is now implemented directly. It covers field arguments, directive arguments, list entries and input object fields, in the operation and in every fragment it can reach. The rule also depends on the default value fix above, so the two cases which allow a nullable variable at a non-null location are now observable end to end.

Five tests in `SkipIncludeSuite` and six in `VariablesSuite` declared a nullable variable at a non-null location. Those documents are invalid under rule 5.8.5, so the declarations are now non-null.

### DIFF
--- a/modules/core/src/main/scala/compiler.scala
+++ b/modules/core/src/main/scala/compiler.scala
@@ -207,6 +207,223 @@ object QueryParser {
 }
 
 /**
+ * Validation of variable usages against the locations where they appear.
+ *
+ * @see
+ *   https://spec.graphql.org/September2025/#sec-All-Variable-Usages-Are-Allowed
+ */
+object VariableUsage {
+
+  /**
+   * Is the use of `varDef` allowed at a location of type `locationType`?
+   *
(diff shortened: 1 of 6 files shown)


milessabin (approved):
LGTM!

PR #910, opened 2026-08-24
Add conformance test suites from September 2025 GraphQL specification

### LINKED ISSUE #26: Extract test suite from spec and reference implementation
There is no official GraphQL test suite, however there are a large number of positive and negative examples in the spec and extensive tests in the reference implementation. We should attempt to extract these out as a conformance/interop test suite.

It would be a good idea to participate in the GraphQL working group to cooperate with any other similar efforts.

hugo-vrijswijk:
Closes #26

Introduces a `ConformanceSuite` and a number of test suites based on examples from the September 2025 GraphQL specification. Roughly, each suite corresponds to a section of the specification, and each test case corresponds to an example or counter-example in the specification text.

The suites cover every example, including the examples which currently fail. Each failing case carries a `.fail` marker. The markers record the current conformance status as a baseline. Out of **221** cases, **31** currently fail. Issue #908 will close 3 of them. Some cases fail because of small bugs. Others fail because a feature is missing, such as allowing `description` on the root schema. Others fail because of missing minor validations.

The suites use these test helpers:

- `parses` to check that a document parses
- `validSchema`/`invalidSchema` to check the validation of a schema
- `validQuery`/`invalidQuery` to check the validation of a query
- `yields` to run a query against a mapping and compare the response
- `yieldsFieldOrder` to run a query against a mapping and compare the order of the response fields

---

Extracting the conformance suite from the spec was done by AI, but all code has been read, reviewed and understood by me.

### DIFF
--- a/modules/core/src/test/scala/conformance/ConformanceSuite.scala
+++ b/modules/core/src/test/scala/conformance/ConformanceSuite.scala
@@ -1,0 +1,190 @@
+// Copyright (c) 2016-2025 Association of Universities for Research in Astronomy, Inc. (AURA)
+// Copyright (c) 2016-2025 Grackle Contributors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//   http://www.apache.org/licenses/LICENSE-2.0
+//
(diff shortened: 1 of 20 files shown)


milessabin (approved):
This is great stuff ... many thanks!

PR #914, opened 2026-08-28
Show keys for "Expected single value" SQL error

hugo-vrijswijk:
Instead of:

```
Expected single value for field 'key1' of type Parent at List(parents), found many
```

Now shows:

```
Expected single value for field 'key1' of type Parent at List(parents), found many (keys: composite_key_parent.key_1 = <many>, composite_key_parent.key_2 = "GBR")
```


milessabin (approved):
LGTM!

PR #917, opened 2026-08-28
Support DBs without LATERAL JOINS

phdoerfler:
This PR makes support for LATERAL JOINS optional. This has been extracted out of the SQLite branch since it can stand on its own. This way it might be easier to review. It moves the two decisions a backend _without lateral joins_ has to make out of `sql-core` and behind abstract hooks.

## Two abstract members on `SqlMappingLike`

`normalizeOffsetLimit` says how a select's `OFFSET`/`LIMIT` is fixed up before it is rendered, and `unionBranchToFragment` how one branch of a union is rendered. `defaultOffsetForSubquery` is gone, since the first of those covers what it did.

I decided to leave those two abstract and not provide a default implementation. I am not aware of any implementation of `SqlMappingLike` outside of this repo. And if there is, all it has to do is implement two small methods.

Every backend in this repo still has lateral joins, so `supportsLateralJoin` is true everywhere and the branches that do without one are not exercised until #864 adds SQLite and #866 adds H2.

## SQL Server's generated SQL changes

SQL Server now emits `OFFSET 0 ROWS` after an `ORDER BY` at the root of a query, where before it did so only in subqueries. The rows and their order are unchanged; only the text is. Keeping the old text would have meant passing a "subquery or root?" boolean through the new hook, so I took the text change.

### DIFF
--- a/modules/doobie-mssql/src/main/scala/DoobieMSSqlMapping.scala
+++ b/modules/doobie-mssql/src/main/scala/DoobieMSSqlMapping.scala
@@ -93,14 +93,17 @@ trait DoobieMSSqlMappingLike[F[_]] extends DoobieMappingLike[F] with SqlMappingL
       // folded to a bare identifier first (issue #342).
       s.toSubquery(s.table.identifier + "_encaps", Laterality.NotLateral)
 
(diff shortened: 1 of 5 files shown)


milessabin (approved):
LGTM!

PR #918, opened 2026-08-29
Encode error locations with `column` instead of `col`

hugo-vrijswijk:
Spec: https://spec.graphql.org/September2025/#sec-Errors.Error-Result-Format

Problems should be encoded with the key `column` instead of `col` to conform to the GraphQL specification. This is a breaking change, but should also bring better error reporting for clients that rely on the GraphQL specification.


milessabin (approved):
LGTM!

PR #919, opened 2026-08-29
Apply the default `@deprecated`` reason when reason is not provided

hugo-vrijswijk:
Reason of `@deprecated` directive should have the default value "No longer supported".


milessabin (approved):
LGTM!

PR #920, opened 2026-08-29
Allow additional optional arguments on interface implementations

hugo-vrijswijk:
Fields on interface implementations can have more arguments, as long as they are default or nullable.


milessabin (approved):
LGTM!

PR #921, opened 2026-08-29
Coerce Int literals to Float in input values

hugo-vrijswijk:
`Value.checkValue` matched only `FloatValue` for `FloatType`. Therefore a query such as `floatField(arg: 123)` failed to compile.

The variable path was correct already, because the JSON extractor matches every number. A Float literal at an Int location is still rejected.

Removes 1 `.fail` from the conformance tests.


milessabin (approved):
LGTM!

PR #924, opened 2026-08-30
Run the NULL-ordering and nullable-parent suites on SQLite and H2

phdoerfler:
Runs the shared suites that SQLite and H2 were not running. SQLite goes from 245 to 256 tests, H2 from 245 to 270, which is parity with Postgres. All green.

## How the gap happened

Each backend once had its own `Doobie<Db>NullOrderingSuite`. In July all of them were deleted as "superseded by the shared SqlNullOrderingMapping/SqlNullOrderingSuite ... This branch will pick that up once it rebases past #870's merge". #870 merged, and the branches did rebase, but that alone wires nothing: the suite has to be named in each backend's `Doobie*Suites.scala`, and `testdata/null-ordering/` needs a schema for the dialect. Neither happened, so NULL ordering has been untested on exactly the backends that hand-roll a `CASE WHEN ... IS NULL` sort prefix to emulate it.

The suites pass unchanged, so no dialect bug was hiding behind this.

## What each one gets

Both get `SqlNullOrderingSuite`, `SqlNullableParentSuite` and `SqlErrorKeysSuite`. The first two need a schema file per dialect; the third reuses the composite-keys dataset and only ever needed wiring.

H2 additionally gets `SqlQualifiedNamesSuite`, `SqlTableNameSuite` and `SqlUnionOrderSuite`, which put their tables in a schema. H2 has `CREATE SCHEMA`; SQLite has nothing to create one with, so it still skips those two datasets.

MySQL (#872) and MariaDB (#873) have the same gap and pick this up when they rebase.

### DIFF
--- a/modules/doobie-h2/src/test/scala/DoobieH2Suites.scala
+++ b/modules/doobie-h2/src/test/scala/DoobieH2Suites.scala
@@ -55,6 +55,10 @@ final class CompositeKeySuite extends DoobieH2DatabaseSuite with SqlCompositeKey
   lazy val mapping = new DoobieH2TestMapping(transactor) with SqlCompositeKeyMapping[IO]
 }
 
+final class ErrorKeysSuite extends DoobieH2DatabaseSuite with SqlErrorKeysSuite {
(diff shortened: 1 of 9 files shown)


milessabin (approved):
LGTM!

PR #930, opened 2026-09-02
Add mapping concepts and query pipeline documentation

phdoerfler:
Adds two documentation pages: `mappings.md`, a tour of the `Mapping` types and the `FieldMapping` building blocks, and `pipeline.md`, a walk through the parse/compile/interpret/complete pipeline they all feed.

## Corrections to the compiler phases page

I also corrected and moved the page from #879. Three things in it were wrong or missing. `Phase` was presented as a trait with a single abstract `transform`, when `transform` has a default implementation which walks the whole query algebra, maintaining the `Elab` context as it descends. Depth and width were described as computed "from the query text alone", when they are computed over the elaborated algebra. And `Introspect` nodes contribute nothing to either measure, so a deeply nested `__schema` query passes whatever limits are configured, which seemed worth saying on a page about defending against expensive queries.

Beyond that the page's wording is untouched: the changes are two inserted paragraphs and two modified lines.

The move to `howto/` is because the page is reference material for installing a phase rather than a step in the tutorial sequence. It carried one sentence referring to "the Star Wars model from the previous chapter", now a link.

## The in-page references are not links

The few "see the pitfall above" references in `mappings.md` were meant to be real anchors, and I could not make them work. mdoc's link checker resolves anchors the way GitHub does, while Laika generates its own slugs, and the two disagree on the same heading: "I/O" ends up as `io` for one and `i-o` for the other, and the spaced dash in a heading like "`EffectField` - an effectful, batchable field" ends up as `---` for one and `-` for the other. An anchor that satisfies Laika leaves mdoc warning on every build, and one that satisfies mdoc fails `laikaSite` outright, since Laika treats an unresolved internal reference as an error rather than a warning. Plain text was the only version that keeps the build clean.

## Disclaimer

This work was co-authored by an LLM.

### DIFF
--- a/README.md
+++ b/README.md
@@ -36,6 +36,9 @@ and [imbus AG](https://www.imbus.de/) over the last five years.
 ## Getting Started
 
 - See the [tutorial](https://typelevel.org/grackle) and accompanying [demo](https://github.com/typelevel/grackle/tree/main/demo/src/main).
+- New to Grackle's mapping model? Start with [Mapping Concepts](https://typelevel.org/grackle/mappings.html) for a guided tour of
+  `ValueMapping`, `GenericMapping`, `CirceMapping`, `SqlMapping`, `ComposedMapping` and friends,
+  and how to pick the right one for your data source.
 - Online Scaladoc is available [here](https://javadoc.io/doc/org.typelevel/grackle-core_2.13).
 - Ask us anything the in **#grackle** channel on the Typelevel [discord server][grackle-dev].
 
@@ -43,25 +46,31 @@ To add Grackle to your project you should add the following to your `build.sbt`,
 
 ```scala
 // Required: Scala 2.13/3.3+
-libraryDependencies += "org.typelevel" %% "grackle-core" % "0.26.0"
+libraryDependencies += "org.typelevel" %% "grackle-core" % "0.30.0"
 
 // Optional: support for in-memory Json backend using circe
-libraryDependencies += "org.typelevel" %% "grackle-circe" % "0.26.0"
+libraryDependencies += "org.typelevel" %% "grackle-circe" % "0.30.0"
 
 // Optional: support for in-memory generic Scala backend using shapeless
-libraryDependencies += "org.typelevel" %% "grackle-generic" % "0.26.0"
+libraryDependencies += "org.typelevel" %% "grackle-generic" % "0.30.0"
 
 // Optional: support for Postgres backend via Doobie (JVM only)
-libraryDependencies += "org.typelevel" %% "grackle-doobie-pg" % "0.26.0"
+libraryDependencies += "org.typelevel" %% "grackle-doobie-pg" % "0.30.0"
 
 // Optional: support for Postgres backend via Skunk
-libraryDependencies += "org.typelevel" %% "grackle-skunk" % "0.26.0"
+libraryDependencies += "org.typelevel" %% "grackle-skunk" % "0.30.0"
 
 // Optional: support for Oracle backend via Doobie (JVM only)
-libraryDependencies += "org.typelevel" %% "grackle-doobie-oracle" % "0.26.0"
+libraryDependencies += "org.typelevel" %% "grackle-doobie-oracle" % "0.30.0"
 
 // Optional: support for SQL Server backend via Doobie (JVM only)
-libraryDependencies += "org.typelevel" %% "grackle-doobie-mssql" % "0.26.0"
+libraryDependencies += "org.typelevel" %% "grackle-doobie-mssql" % "0.30.0"
+
+// Optional: support for H2 backend via Doobie (JVM only)
+libraryDependencies += "org.typelevel" %% "grackle-doobie-h2" % "0.30.0"
+
+// Optional: support for SQLite backend via Doobie (JVM only)
+libraryDependencies += "org.typelevel" %% "grackle-doobie-sqlite" % "0.30.0"
 ```
 
 ## Running tests for database backed mappings
--- a/docs/howto/compiler-phases.md
+++ b/docs/howto/compiler-phases.md
@@ -26,6 +26,10 @@ in the `Elab` monad. `Elab` gives phases access to the schema, the current conte
 definitions, and allows compilation to be aborted with one or more GraphQL errors via `Elab.failure`. A phase which
 fails prevents the query from executing at all — the client receives an error response with no data.
 
+`transform` is not abstract: the inherited implementation walks the whole query algebra, maintaining the `Elab`
+context as it descends, so a phase overrides it for the node types it cares about and delegates the rest to
+`super.transform`.
+
 This makes phases a natural place to enforce global policies on incoming queries. Grackle provides one such policy
 phase out of the box: `QuerySizeValidator`.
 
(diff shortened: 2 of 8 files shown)


docs/howto/compiler-phases.md
```diff
@@ -84,8 +97,12 @@ exceeding both limits at once is reported as `"Query is too complex"`.
 
 ## Limitations
 
-Depth and width are syntactic measures: they are computed from the query text alone and know nothing about the size
+Depth and width are structural measures: they are computed from the query's shape and know nothing about the size
```
milessabin, line 91: I'm not convinced this is an improvement. The key contrast is between the query text, which is syntax, and data-independent, and the result size, which _is_ data dependent.

docs/howto/compiler-phases.md
```diff
@@ -84,8 +97,12 @@ exceeding both limits at once is reported as `"Query is too complex"`.
...
 of the underlying data. In particular, width does not account for list sizes — a field yielding a thousand elements
 contributes to the width just once. `QuerySizeValidator` is therefore a coarse first line of defence rather than a
 complete cost model. Guarding against expensive list expansions requires taking field cardinalities and arguments
 into account, which can be implemented as a custom phase following the same pattern.
+
+Introspection is not measured at all: an `Introspect` node contributes nothing to either figure, so a deeply
```
milessabin, line 97: Not to say that this is inaccurate, but it might actually be a problem 🤔

docs/index.md
```diff
@@ -56,6 +59,12 @@ libraryDependencies += "org.typelevel" %% "grackle-doobie-oracle" % "@VERSION@"
 
 // Optional: support for SQL Server backend via Doobie (JVM only)
 libraryDependencies += "org.typelevel" %% "grackle-doobie-mssql" % "@VERSION@"
+
+// Optional: support for H2 backend via Doobie (JVM only)
+libraryDependencies += "org.typelevel" %% "grackle-doobie-h2" % "@VERSION@"
+
+// Optional: support for SQLite backend via Doobie (JVM only)
+libraryDependencies += "org.typelevel" %% "grackle-doobie-sqlite" % "@VERSION@"
```
milessabin, line 67: More to add now 😄

docs/pipeline.md
```diff
@@ -0,0 +1,171 @@
...
+backend replaces the compiler or the interpreter - and a `Mapping` plugs into it at the handful of
+points highlighted below. That is why we can mix `ValueMapping`, `SqlMapping`, effects and
+`ComposedMapping` freely: they differ mainly in how they build cursors, and everything either side
+of that is shared. `ValueMapping`, `GenericMapping`, `CirceMapping` and `ComposedMapping` each
+override `mkCursorForMappedField`, falling through to `super` for the cases they do not handle;
+`SqlMapping` instead overrides `defaultRootCursor`, because it materialises a whole subtree from a
+single query rather than a cursor per field. `SelectElaborator`, `RootEffect`/`RootStream` and
+`EffectHandler` are orthogonal to that choice: they work under any of them.
+
+```mermaid
```
milessabin, line 14: Diagram is excellent! 💯

docs/pipeline.md
```diff
@@ -0,0 +1,171 @@
+# How a query flows through Grackle
+
+Grackle is a compiler/interpreter: a request is parsed once, elaborated into an executable query
```
milessabin, line 3: Reads a bit LLM-ish, but not too bad.

docs/pipeline.md
```diff
@@ -0,0 +1,171 @@
...
+- **Elaboration is where "arguments become behavior".** A raw `UntypedSelect` does not know what
+  `character(id: "1000")` *means* - `mapping.selectElaborator` is what turns that into a `Filter`
+  or `Unique` node (or an `Elab.env` binding) parameterized by the argument value. This is the
+  piece that is most specific to the data source, and it is exactly what `SelectElaborator.apply`
+  (see the [tutorial](tutorial/intro.md)) lets us supply.
+- **Batching happens after the fact, not during the walk.** `runValue`/`runFields` build the
+  result tree eagerly, but `Component` and `Effect` nodes do not run immediately - they produce a
+  deferred `ProtoJson` marker. `completeAll` is what gathers *all* the deferred markers across the
+  whole response, groups them by which `Mapping`/`EffectHandler` is responsible, and runs each
+  group exactly once. That is the mechanism behind `SqlMapping`'s single-query-per-level joins and
```
milessabin, line 112: This is correct as far as `EffectField` is concerned, but not wrt `SqlMapping` (although it was true a very very long time ago). Eliminating N+1 in the SQL part of the process is handled by the compilation to SQL, which now I think about it is missing from the diagram.

docs/pipeline.md
```diff
@@ -0,0 +1,171 @@
...
+  `compilerPhases`: by default `selectElaborator`, `componentElaborator`, `effectElaborator`, but a
+  mapping can append phases of its own to enforce global policy on incoming queries. Grackle ships
+  one such phase, `QuerySizeValidator`; see
+  [Compiler Phases](howto/compiler-phases.md) for how to write and install our own.
+- **`ComposedMapping` is this whole diagram, nested.** A `Component` node's "hand the subtree to
+  another Mapping" step (`I6` above) means that sub-mapping runs its *own* copy of stages ②-④ on
+  that subtree, with its own schema and its own `Env`. Stitching mappings together is possible
+  precisely because every `Mapping` implements the same pipeline independently.
+
+## What a phase is, and what it sees
```
milessabin, line 125: It hate this "What X Y's, and what it Zs" construction ... blatant Claude.

docs/pipeline.md
```diff
@@ -0,0 +1,171 @@
...
+
+Stage ② is a fold over a list of `QueryCompiler.Phase`s, run once per operation:
+
+```scala
+allPhases.foldLeftM(op.query) { (acc, phase) =>
+  phase.transformFragments *> phase.transform(acc)
+}
+```
+
+`Phase` has two entry points, and neither is abstract:
```
milessabin, line 135: Not super keen on this section ... needs more human review.

docs/pipeline.md
```diff
@@ -0,0 +1,171 @@
...
+  `ElabState`, not in the query tree, so `transform` never reaches inside them; a phase which has
+  to rewrite fragment bodies overrides `transformFragments` and rewrites the map through
+  `Elab.transformFragments`. Only phases running before fragments are inlined need this -
+  `IntrospectionElaborator` is the one built-in that does, because `__schema` can be selected
+  inside a fragment.
+
+`transformSelect` and `validateSubselection` are the seams inside the default `transform`, for
+hooking field descent rather than replacing the walk.
+
+Where a phase sits in the list decides which nodes it can match on at all:
```
milessabin, line 153: This is good, but the motivation is missing ... most users don't need to add phases _at all_. I think it needs to be clear that these details are only relevant when you need to extend the framework (eg. adding additional pre/post processing, validation, delegation etc. etc.).

README.md
```diff
@@ -36,32 +36,41 @@ and [imbus AG](https://www.imbus.de/) over the last five years.
...
 
 // Optional: support for SQL Server backend via Doobie (JVM only)
-libraryDependencies += "org.typelevel" %% "grackle-doobie-mssql" % "0.26.0"
+libraryDependencies += "org.typelevel" %% "grackle-doobie-mssql" % "0.30.0"
+
+// Optional: support for H2 backend via Doobie (JVM only)
+libraryDependencies += "org.typelevel" %% "grackle-doobie-h2" % "0.30.0"
+
+// Optional: support for SQLite backend via Doobie (JVM only)
+libraryDependencies += "org.typelevel" %% "grackle-doobie-sqlite" % "0.30.0"
```
milessabin, line 73: MOAR!!!

docs/pipeline.md
milessabin: Although see comment below about missing SQL compilation.

milessabin (changes requested):
Not bad, but needs more humanising ...

PR #937, opened 2026-09-04
Factor out common db test settings and force UTC for all

milessabin:
The test failures with non-UTC timezones issue showed up for me locally for MariaDB ... IOW @phdoerfler was right in https://github.com/typelevel/grackle/pull/862. This factors out some common DB test settings and applies the timezone fix globally.


hugo-vrijswijk (approved):
LGTM

PR #940, opened 2026-09-08
Make MappingValidator stack safe

rpiaggio:
Use `Eval` instead of `Id` to keep `MappingValidator` stack safe.

In a schema with 30+k lines this got to be an issue.


hugo-vrijswijk (approved):
LGTM! Apart from a scalafmt issue

PR #941, opened 2026-09-08
Add `location` to error messages

hugo-vrijswijk:
Adds a `location` field `Ast.Selection.Field` to hold the line and column of the selection. That can then be used to add a `locations` field to error messages.

Most changes are in tests with added `location` information to the error messages. The other option would be to strip location information from the expected error messages, but that would make it harder to verify that the location information is actually present.

Fixes 2 conformance tests.

### DIFF
--- a/docs/tutorial/in-memory-model.md
+++ b/docs/tutorial/in-memory-model.md
@@ -209,9 +209,10 @@ Grackle's query algebra consists of the following elements,
 case class UntypedSelect(
   name: String, alias: Option[String],
   args: List[Binding], directives: List[Directive],
-  child: Query
+  child: Query,
+  location: Option[(Int, Int)] = None
 )
-case class Select(name: String, alias: Option[String], child: Query)
+case class Select(name: String, alias: Option[String], child: Query, location: Option[(Int, Int)] = None)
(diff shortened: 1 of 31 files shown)


docs/tutorial/in-memory-model.md
```diff
@@ -209,9 +209,10 @@ Grackle's query algebra consists of the following elements,
 case class UntypedSelect(
   name: String, alias: Option[String],
   args: List[Binding], directives: List[Directive],
-  child: Query
+  child: Query,
+  location: Option[(Int, Int)] = None
 )
-case class Select(name: String, alias: Option[String], child: Query)
+case class Select(name: String, alias: Option[String], child: Query, location: Option[(Int, Int)] = None)
```
hugo-vrijswijk, line 215: This is also a source-breaking change for any users that pattern-match on `Select(n, a, c)`, but I don't know if there is a way around

milessabin (approved):
LGTM!

docs/tutorial/in-memory-model.md
milessabin: I'm not too concerned about this: `SelectElatorator` should protect most uses from this.

PR #942, opened 2026-09-09
Fix Binding Value rendering

hugo-vrijswijk:
Now renders the value instead of the case-class toString

Also fixes string rendering for values with escape codes like newlines or quotes.


milessabin (approved):
LGTM!

PR #947, opened 2026-09-11
Carve out parser as a separate artifact

rpiaggio:
This allows us to use the parser in clue for client-side validation of queries.


build.sbt
```diff
@@ -263,17 +264,37 @@ lazy val modules: List[CompositeProject] = List(
...
+    tlVersionIntroduced := Map("2.13" -> "0.31.0", "3" -> "0.31.0"),
+    libraryDependencies ++= Seq(
+      "org.typelevel" %%% "cats-parse" % catsParseVersion,
+      "org.typelevel" %%% "cats-core" % catsVersion,
+      "io.circe" %%% "circe-core" % circeVersion
+    )
+  )
+  .jsSettings(scalaJSLinkerConfig ~= (_.withModuleKind(ModuleKind.CommonJSModule)))
+  .nativeSettings(nativeSettings)
+  .nativeSettings(tlVersionIntroduced := Map("2.13" -> "0.31.0", "3" -> "0.31.0"))
```
hugo-vrijswijk: Isn't this already defined in the cross-platform settings?

rpiaggio: It is — thanks. The extra line was only there to undo `nativeSettings`, which sets
`tlVersionIntroduced` to 0.26.1 (when Native moved to 0.5); that's wrong for a module
that's new in 0.31.0 on every platform. Simpler not to apply `nativeSettings` here at
all, so both lines are gone and Native picks up 0.31.0 from the cross-platform settings.

armanbilge: Actually, I think `tlVersionIntroduced` shouldn't be necessary anymore at all.

> that's wrong for a module that's new in 0.31.0 on every platform

Indeed, but it won't be checking for bincompat for any previous version anyway.

